PR #162, opened 2026-03-15
feat: ai translation service

cungminh2710:
## Checklist

<h3>Greptile Summary</h3>

This PR introduces a new async AI translation service (`translationsvc`) with an in-memory job state machine, outbox-based segment dispatch, and both inline-string and artifact-file translation modes. It replaces the previous stub backend with a working in-memory implementation wired through the `tmsgrpc.StubBackend`, adds REST endpoints for the translation job lifecycle (`POST`, `GET`, `LIST`, `CANCEL`), defines the gRPC proto contract, and ships a database migration schema — a substantial and well-structured foundation for the translation workflow.

Key findings from the review:

- **TOCTOU race in `FinalizeJob` (artifact mode)** — `finalizeStatus` is captured under the first lock and applied after an I/O-induced re-lock. A concurrent `CancelTranslationJob` between the two lock acquisitions causes the job to be saved as `"completed"` instead of `"canceled"`, silently discarding the cancellation. `finalizeStatus` must be re-derived from the fresh `job` after the second `s.mu.Lock()`.
- **`cancel_requested` maps to `"canceled"` before finalization** — The HTTP cancel endpoint returns `status: "canceled"` immediately, but `FinalizeJob` hasn't yet run, so `inlineResult` / `outputArtifactUri` are still empty. Clients that stop polling on seeing a terminal status will miss partial inline results written at finalization. Consider mapping `cancel_requested` → `"running"` and only returning `"canceled"` once the job reaches the fully finalized `StatusCanceled`.
- **`limit` query parameter accepts negative values** — `strconv.Atoi` parses `limit=-1` without error; the service silently treats it as "no limit" and returns all records. Explicit validation is needed.
- **Unknown job sub-path action returns 405 instead of 404** — `POST /v1/translation-jobs/{id}:anything-unknown` falls through to `w.WriteHeader(http.StatusMethodNotAllowed)`, which is semantically wrong when the method is valid but the action doesn't exist.
- **Test coverage gap** — `service_test.go` reimplements worker logic with locally-defined types instead of exercising `translationworker.Service.HandleExecute` directly, so changes to the worker's interface won't be caught by this suite.

<h3>Confidence Score: 2/5</h3>

- Not safe to merge — a TOCTOU race in the artifact finalization path can silently discard job cancellations, marking canceled jobs as completed.
- The TOCTOU race in FinalizeJob for artifact-mode jobs is a clear correctness bug: a concurrent CancelTranslationJob between the two lock acquisitions causes `finalizeStatus` to be applied from stale state, so the job is persisted as "completed" instead of "canceled". Several other pre-existing issues (noted in earlier review threads) remain partially addressed. The `mapStatus` fix for internal status leakage was applied but introduces a new premature-terminal-status problem for `cancel_requested`. The overall architecture is sound and well-tested for the happy path, but the artifact finalization race needs to be resolved before production use.
- services/translationsvc/service.go (TOCTOU race in FinalizeJob artifact path), pkg/client/tmsgrpc/client.go (premature canceled status for cancel_requested), apps/api-gateway/internal/http/server.go (limit validation and 405 vs 404 for unknown actions)

<details open><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| services/translationsvc/service.go | Core in-memory translation job service with outbox pattern; contains a TOCTOU race in FinalizeJob artifact mode where a stale `finalizeStatus` can incorrectly mark a concurrently-canceled job as "completed". |
| apps/api-gateway/internal/http/server.go | HTTP route registration refactored to translation job endpoints; `parseListFilter` accepts negative `limit` values (silently becomes no-limit), and unknown job actions incorrectly return 405 instead of 404. |
| pkg/client/tmsgrpc/client.go | StubBackend replaced with an in-memory translationsvc-backed implementation; `mapStatus` correctly normalises internal states but maps `cancel_requested` to "canceled" before finalization, causing clients to see a terminal status before results are available. |
| db/migrations/202603150001_translation_jobs.sql | Well-structured schema for translation jobs, segments, attempts, and artifacts; partial index on dedupe correctly only enforces uniqueness when idempotency_key is present. |
| services/translationsvc/service_test.go | Good coverage of inline and artifact job lifecycle, idempotency, cancellation, and cursor pagination; test helper `handleExecute` reimplements worker logic with locally-defined types rather than exercising `translationworker.Service` directly. |
| services/translationworker/service.go | Thin worker that delegates to a `TranslationService` interface and an `Executor`; depends on an interface rather than a concrete type, avoiding the circular-dependency issue that would arise from a concrete reference. |
| domains/translation/model.go | Domain models and status constants are well-defined and comprehensive; internal lifecycle states (finalize_queued, cancel_requested) are cleanly separated from the public API-facing proto enum. |
| api/proto/hyperlocalise/tms/translation/v1/translation.proto | Proto contract for TranslationService is clean and minimal; intentionally omits internal fields (mode, itemCount, progress) that are present in the REST schema — documented divergence should be noted in the mapJob adapter. |
| services/translationsvc/support.go | In-memory dispatcher and artifact store test helpers; exported from the main package (not a testutil package) but clearly marked with TODO comments for replacement — acceptable for a stub phase. |

</details>

</details>

<details><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant Client
    participant HTTP as HTTP Server
    participant Backend as StubBackend
    participant TMS as TranslationSvc
    participant Publisher
    participant Worker as TranslationWorker
    participant Executor as AI Executor

    Client->>HTTP: POST /v1/translation-jobs
    HTTP->>Backend: CreateTranslationJob
    Backend->>TMS: CreateTranslationJob(input)
    TMS->>TMS: createJobLocked - plan segments
    TMS->>TMS: enqueueExecuteLocked per segment
    TMS->>TMS: FlushOutbox
    TMS->>Publisher: PublishExecute(msg)
    TMS-->>Backend: Job status queued
    Backend-->>HTTP: TranslationJob status pending
    HTTP-->>Client: 202 Accepted

    Worker->>TMS: StartSegmentAttempt(msg)
    TMS-->>Worker: enriched msg and attempt
    Worker->>Executor: Translate(req)
    Executor-->>Worker: Response with translated text
    Worker->>TMS: CompleteSegmentAttempt(segmentID, text, latency)
    TMS->>TMS: recomputeJob - if all terminal set finalize_queued
    TMS->>TMS: enqueueFinalizeLocked
    TMS->>TMS: FlushOutbox
    TMS->>Publisher: PublishFinalize(jobID)

    TMS->>TMS: FinalizeJob(id)
    alt inline mode
        TMS->>TMS: collect InlineOutput and set StatusCompleted
    else artifact mode
        TMS->>TMS: marshal payload under lock then Unlock
        TMS->>TMS: artifactStore.Put(key, payload)
        TMS->>TMS: re-Lock then set OutputArtifactURI and Status
    end

    Client->>HTTP: GET /v1/translation-jobs/{id}
    HTTP->>Backend: GetTranslationJob(id)
    Backend->>TMS: GetJob(id)
    TMS-->>Backend: Job status completed with result
    Backend-->>HTTP: TranslationJob status succeeded
    HTTP-->>Client: 200 OK with inlineResult or artifactResult
```
</details>

<details open><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `apps/api-gateway/internal/http/server.go`, line 45-108 ([link](https://github.com/quiet-circles/hyperlocalise/blob/64d1ffc7085ac8a79c1eb3d6337f56b5b20092e1/apps/api-gateway/internal/http/server.go#L45-L108)) 

   **Previously registered routes silently dropped**

   This PR removes the HTTP handlers for `/v1/projects`, `/v1/resources`, `/v1/translation-memory`, `/v1/glossaries`, and `/v1/workflows` from `registerRoutes`. There is no mention of this in the PR description and the `Backend` interface has dropped the corresponding methods entirely. Any client currently calling these paths will receive `404` rather than a `405` or a deprecation notice.

   If this is intentional (e.g. the plan is to re-introduce these routes under the new service architecture in a follow-up), consider at minimum registering a stub handler that returns `501 Not Implemented` with an informative message, or documenting the removal in the PR description so reviewers can assess the blast radius.

   <details><summary>Prompt To Fix With AI</summary>

   `````markdown
   This is a comment left during a code review.
   Path: apps/api-gateway/internal/http/server.go
   Line: 45-108

   Comment:
   **Previously registered routes silently dropped**

   This PR removes the HTTP handlers for `/v1/projects`, `/v1/resources`, `/v1/translation-memory`, `/v1/glossaries`, and `/v1/workflows` from `registerRoutes`. There is no mention of this in the PR description and the `Backend` interface has dropped the corresponding methods entirely. Any client currently calling these paths will receive `404` rather than a `405` or a deprecation notice.

   If this is intentional (e.g. the plan is to re-introduce these routes under the new service architecture in a follow-up), consider at minimum registering a stub handler that returns `501 Not Implemented` with an informative message, or documenting the removal in the PR description so reviewers can assess the blast radius.

   How can I resolve this? If you propose a fix, please make it concise.
   `````
   </details>
</details>

<details><summary>Prompt To Fix All With AI</summary>

`````markdown
This is a comment left during a code review.
Path: services/translationsvc/service.go
Line: 388-395

Comment:
**TOCTOU race: stale `finalizeStatus` applied after re-acquiring lock**

`finalizeStatus` is computed from the job's state under the *first* lock (around line 325–330). The mutex is then released for the `artifactStore.Put` I/O call. Between that unlock and the second `s.mu.Lock()`, a concurrent `CancelTranslationJob` can change `job.Status` from `running`/`queued` to `cancel_requested`. When the second lock is acquired, the re-read `job` now has `cancel_requested` status, but the re-lock guard only bails out for fully terminal jobs (`isTerminalJob` returns false for `cancel_requested`), so the stale `finalizeStatus = StatusCompleted` is applied, silently discarding the cancellation.

The `finalizeStatus` must be re-derived under the second lock using the freshly read `job` and segments:

```go
s.mu.Lock()
defer s.mu.Unlock()

job, ok = s.jobs[id]
if !ok {
    return translation.Job{}, ErrNotFound
}
if isTerminalJob(job.Status) && job.OutputArtifactURI != "" {
    return cloneJob(job), nil
}

// Re-derive finalizeStatus from the current (post-I/O) job state.
segments = s.segments[id]
finalizeStatus = translation.StatusCompleted
switch {
case job.Status == translation.StatusCancelRequested:
    finalizeStatus = translation.StatusCanceled
case hasFailedSegments(segments):
    finalizeStatus = translation.StatusFailed
}

job.Status = finalizeStatus
job.OutputArtifactURI = outputURI
```

How can I resolve this? If you propose a fix, please make it concise.

---

This is a comment left during a code review.
Path: apps/api-gateway/internal/http/server.go
Line: 156-162

Comment:
**No validation for negative or zero `limit` values**

`strconv.Atoi` successfully parses `limit=-1` and `limit=0` without error, and both values flow into the service unchanged. In `ListTranslationJobs`, the guard `if filter.Limit > 0` means a negative or zero limit is treated as "no limit" (all jobs are returned). A caller sending `limit=0` expecting an empty page, or `limit=-5` as an error case, would silently receive all records — a potential DoS vector if the job store grows large.

Consider rejecting non-positive values explicitly:

```suggestion
		if raw := strings.TrimSpace(query.Get("limit")); raw != "" {
			limit, err := strconv.Atoi(raw)
			if err != nil {
				return tmsgrpc.TranslationJobFilter{}, errors.New("limit must be an integer")
			}
			if limit <= 0 {
				return tmsgrpc.TranslationJobFilter{}, errors.New("limit must be a positive integer")
			}
			filter.Limit = limit
		}
```

How can I resolve this? If you propose a fix, please make it concise.

---

This is a comment left during a code review.
Path: apps/api-gateway/internal/http/server.go
Line: 95-107

Comment:
**Unknown action path returns 405 Method Not Allowed instead of 404 Not Found**

When a client sends `POST /v1/translation-jobs/{id}:unknown-action`, `parseJobPath` returns `action = "unknown-action"`. The switch falls through to `default: w.WriteHeader(http.StatusMethodNotAllowed)`. This is semantically wrong — the HTTP method (`POST`) is valid, but the resource path or action simply doesn't exist. RFC 9110 designates 404 for that case.

Consider splitting the default cases to distinguish "known action, wrong method" from "unknown action":

```go
switch {
case action == "" && r.Method == http.MethodGet:
    // ...
case action == "cancel" && r.Method == http.MethodPost:
    // ...
case action == "cancel":
    w.WriteHeader(http.StatusMethodNotAllowed)
default:
    w.WriteHeader(http.StatusNotFound)
}
```

How can I resolve this? If you propose a fix, please make it concise.

---

This is a comment left during a code review.
Path: pkg/client/tmsgrpc/client.go
Line: 183-186

Comment:
**`cancel_requested` mapped to "canceled" before job is finalized**

`mapStatus` maps `cancel_requested` to `"canceled"` immediately when `CancelTranslationJob` is called. From the client's perspective, `"canceled"` looks like a terminal state, but `FinalizeJob` hasn't run yet — `inlineResult` will be `null` for inline jobs that have partial completions, and `outputArtifactUri` will be empty for artifact jobs.

A client that stops polling upon seeing `"canceled"` will never observe the partial `inlineResult` that `FinalizeJob` writes once all in-flight segments settle. A safer mapping would return `"running"` for `cancel_requested` and only return `"canceled"` once the job reaches the final `StatusCanceled` state (i.e., after finalization):

```go
case translation.StatusCancelRequested:
    return "running" // cancellation is in progress, not yet finalized
case translation.StatusCanceled:
    return "canceled"
```

This ensures clients can rely on the `"canceled"` status meaning the job is fully finalized with whatever partial results are available.

How can I resolve this? If you propose a fix, please make it concise.
`````

</details>

<sub>Last reviewed commit: a40fba4</sub>

> Greptile also left **3 inline comments** on this PR.

### CODE REVIEW

File: services/translationworker/service.go [hunk 1]
```diff
@@ -0,0 +1,81 @@
...
+		SourceText:      segmentMsg.SourceText,
+		Context:         segmentMsg.Context,
+		SourceLocale:    segmentMsg.SourceLocale,
+		TargetLocale:    segmentMsg.TargetLocale,
+		Attempt:         segmentMsg.Attempt,
+		ProviderProfile: segmentMsg.ProviderProfileID,
+	})
+	latency := time.Since(started)
+	if err != nil {
+		_, failErr := s.translationService.FailSegmentAttempt(ctx, segmentMsg.SegmentID, "provider_error", err.Error(), latency)
```
MuenYu: magic error code

PR #525, opened 2026-05-18
⚡ Bolt: optimize segment processing hot paths

cungminh2710:
This PR optimizes several hot paths in the i18n segment processing logic used across HTML and Markdown parsers.

### Changes:
- **Fast HTML Tag Detection:** `containsHTMLTag` now uses a `strings.Contains(s, "<")` fast-path. Benchmarks show a ~5x speedup for segments without markup.
- **Efficient Placeholder Expansion:** Added a `strings.ReplaceAll` fast-path for segments with exactly one placeholder, providing a ~10x speedup by avoiding `strings.Replacer` initialization overhead.
- **Allocation Reduction:** `isTranslatableChunk` now iterates over the string directly instead of calling `strings.TrimSpace`, avoiding a potential heap allocation.
- **Redundant Sort Removal:** Removed unnecessary sorting of sentinel tokens in `expandHTMLPlaceholders`. Since tokens are fixed-length and non-colliding, the sort was adding $O(N \log N)$ overhead and an extra allocation for no functional benefit.

### Impact:
These optimizations significantly reduce the CPU and memory footprint of document marshalling, especially for files with many small translatable segments.

### Verification:
- Ran micro-benchmarks confirming the speedups.
- Verified all existing parser and translator tests pass with `go test ./internal/i18n/...`.

---
*PR created automatically by Jules for task [16943343225582222609](https://jules.google.com/task/16943343225582222609) started by @cungminh2710*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -11,3 +11,7 @@
 ## 2026-05-17 - Caching strings.Replacer for multi-segment rendering
 **Learning:** Rebuilding a `strings.Replacer` for every segment in a document rendering loop is expensive due to trie construction. In this codebase, Liquid placeholders are document-wide, meaning the same replacer can be reused across all segments in a `liquidDocument`.
 **Action:** Move `strings.Replacer` initialization to the document level (`parseLiquidDocument`) and cache it in the document struct to achieve ~85% faster rendering and ~95% fewer allocations.
+
+## 2026-05-20 - Optimizing segment processing hot paths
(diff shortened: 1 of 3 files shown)


cungminh2710:
@jules The logic changes are correct: the \x1f-terminated sentinel format truly prevents prefix collisions, making the removed sort unnecessary. The isTranslatableChunk rewrite is semantically identical. The only rough edge is that the expandHTMLPlaceholders docstring and one inline comment are now misleading about token properties, which could confuse future contributors reasoning about placeholder safety.

The function-level docstring and inline comment in html_parser.go's expandHTMLPlaceholders need updating to reflect why sorting is safe to omit.

PR #650, opened 2026-05-22
feat(web): show synced TMS files in unified Files UI (HL-366)

cungminh2710:
## What changed

Extends the existing Files experience so native Hyperlocalise repository sources and synced TMS provider files/keys appear in one UI.

- **Org Files page** (`/org/[slug]/files`): unified list grouped by project with search, provider, project, source type (repository vs provider / file vs key), locale, and sync state filters.
- **Project Files page**: same filter bar and richer metadata (provider badge, origin, resource type, sync state, locale readiness summary, last sync).
- **API**: `GET /api/orgs/:organizationSlug/workspace-files` for org-wide browsing; project `GET .../files` gains query filters and `lastSyncedAt` on provider payloads.
- **Shared logic**: `src/lib/projects/project-files.ts` centralizes listing, merging repository + `external_tms_files`, and server-side filtering.

Selecting a file links to the existing project file detail flow via `?sourcePath=...`.

Resolves HL-366.

## How to test

1. Start Postgres (`docker compose up -d`) and apply migrations (`vp run db:migrate` in `apps/hyperlocalise-web`).
2. Seed or sync provider files for a project (Phrase/Crowdin fixtures in existing tests).
3. Open `/org/{slug}/files` — confirm native and provider items appear with filters and project grouping.
4. Open `/org/{slug}/projects/{id}/files` — confirm filters, badges, and deep link selection via `sourcePath`.
5. Run checks:
   - `cd apps/hyperlocalise-web && vp check --fix`
   - `vp test src/lib/projects/project-files.test.ts`
   - `vp test src/api/routes/workspace-files/workspace-files.test.ts`
   - `vp test src/api/routes/project/project.test.ts` (requires DB)

## Checklist

- [x] I ran `vp check --fix` locally
- [x] Unit test for `filterProjectFiles` passes
- [ ] Integration API tests pass in CI (require Postgres)

Linear Issue: [HL-366](https://linear.app/hyperlocalise/issue/HL-366/show-synced-tms-files-and-keys-in-existing-files-page)

<div><a href="https://cursor.com/agents/bc-ff58f754-eb92-4f9e-86c1-ef12af347ba7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ff58f754-eb92-4f9e-86c1-ef12af347ba7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/hyperlocalise-web/src/api/app.ts
+++ b/apps/hyperlocalise-web/src/api/app.ts
@@ -30,6 +30,7 @@ import { createResendWebhookRoutes } from "./routes/resend-webhook/resend-webhoo
 import { createSlackOAuthRoutes } from "./routes/slack-oauth/slack-oauth.route";
 import { createSlackWebhookRoutes } from "./routes/slack-webhook/slack-webhook.route";
 import { createFileRoutes } from "./routes/file/file.route";
+import { createWorkspaceFilesRoutes } from "./routes/workspace-files/workspace-files.route";
 import { createExternalTmsProviderCredentialRoutes } from "./routes/external-tms-provider-credential/external-tms-provider-credential.route";
 import { createTeamRoutes } from "./routes/team/team.route";
 import { workosWebhookRoutes } from "./routes/workos-webhook/workos-webhook.route";
@@ -110,6 +111,7 @@ function createOrgScopedAppRoutes(
     .route("/agent-slack", createAgentSlackRoutes())
     .route("/teams", createTeamRoutes())
     .route("/files", createFileRoutes({ fileStorageAdapter: options.fileStorageAdapter }))
+    .route("/workspace-files", createWorkspaceFilesRoutes())
     .route("/conversations", createConversationRoutes())
     .route(
       "/chat-requests",
--- a/apps/hyperlocalise-web/src/api/routes/project/project.route.ts
+++ b/apps/hyperlocalise-web/src/api/routes/project/project.route.ts
@@ -28,7 +28,7 @@ import {
   pullExternalTmsTaskContent,
   pushExternalTmsTranslations,
 } from "@/lib/providers/external-tms-content-sync";
-import { filterProjectFiles, listProjectFilesForProject } from "@/lib/projects/project-files";
+import { listFilteredProjectFiles } from "@/lib/projects/project-files";
 import type { ExternalTmsResourceType } from "@/lib/providers/organization-external-tms-files";
 import { bufferFromStream } from "@/lib/streams";
 import { inferSupportedFileTranslationFileFormat } from "@/lib/translation/file-formats";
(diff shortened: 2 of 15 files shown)


cungminh2710:

---

### Issue 1 of 5
apps/hyperlocalise-web/src/app/(authenticated)/org/[organizationSlug]/projects/[projectId]/files/_components/project-files-page-content.tsx:1198-1208
**URL deep-link `sourcePath` cleared before files load**

The second `useEffect` runs immediately after mount with `files = []` (the query hasn't resolved yet). Because no file in the empty array matches `selectedPath`, it resets `selectedPath` to `undefined`. When files eventually load, the guard `if (selectedPath && ...)` is false and nothing restores the path — so any link navigating to `?sourcePath=some/path` silently drops the selection.

### Issue 2 of 5
apps/hyperlocalise-web/src/lib/projects/project-files.ts:1552-1557
**DB-level limit applied before server-side filtering, producing silently incomplete results**

`listExternalTmsFilesForProject` is called with `input.limit` (typically 500) before any search, locale, or syncState filters are applied. `filterProjectFiles` is then run over those 500 results. If a project has more than 500 provider files, any search term or filter will only evaluate the first 500 rows from the database — matching files beyond that window are silently dropped.

### Issue 3 of 5
apps/hyperlocalise-web/src/app/(authenticated)/org/[organizationSlug]/_components/workspace-files-shared.tsx:829-863
**`useClientFileFilters` is exported but never called**

Neither `files-page-content.tsx` nor `project-files-page-content.tsx` imports or calls this hook — both pages pass `filters` into the server query key and let the API filter. The hook is dead code as shipped.

### Issue 4 of 5
apps/hyperlocalise-web/src/lib/projects/project-files.ts:1427-1430
**Duplicate `WorkspaceFileRecord` type definition**

`WorkspaceFileRecord` is declared here as `ProjectFileRecord & { projectId: string; projectName: string }` and again in `project.schema.ts` via `workspaceFileRecordSchema.extend(...)`. A future schema change would need to be updated in both places.

### Issue 5 of 5
apps/hyperlocalise-web/src/lib/projects/project-files.ts:1691-1698
**Per-project pre-limit skews results when combined with search filters**

`perProjectLimit` distributes the total budget evenly across all projects (`ceil(limit / numProjects)`). With 20 projects and `limit=500`, each project is capped at 25 files before filtering — a project that has 40 matches will only return 25.

cungminh2710:
@cursoragent  Fix the following 2 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 2
apps/hyperlocalise-web/src/lib/projects/project-files.ts:319-338
**Unbounded `Promise.all` fan-out per org project**

`listWorkspaceFiles` fires one `listProjectFilesForProject` call per project concurrently via `Promise.all` with no concurrency cap. Each call makes 3–5 DB round-trips (versions subquery, provider files, jobs subquery when versions exist). An org with 50 projects therefore issues up to 250 simultaneous queries, which can exhaust the connection pool and cause cascading timeouts for other requests. Consider batching with a concurrency limiter (e.g. `p-limit`) or switching to a single multi-project query for the versions/jobs data.

### Issue 2 of 2
apps/hyperlocalise-web/src/app/(authenticated)/org/[organizationSlug]/files/_components/files-page-content.tsx:56-60
**Locale dropdown options derived from already-filtered response**

`localeOptions` is computed from `files`, which is the API response after the current filters have already been applied. If the user picks `origin: "repository"` and repo files carry no locale data, `localeOptions` becomes empty — but any previously-selected locale value remains active in the filter state (the `Select` trigger shows it, the dropdown only offers "All locales"). The user then sees an empty result set with no obvious way to identify that a stale locale filter is still active. The same pattern is present in `project-files-page-content.tsx`.

A straightforward mitigation is to populate locale options from a separate unfiltered call or reset `locale` to `"all"` whenever `localeOptions` no longer contains the current selection.

PR #831, opened 2026-05-31
⚡ Bolt: optimize XLIFF parsing and marshaling

cungminh2710:
Optimized XLIFF processing in `internal/i18n/translationfileparser`. 

Key improvements:
- **XLIFF Parsing**: Switched from a token-by-token re-encoding approach to raw slicing of the input buffer using `decoder.InputOffset()`. This drastically reduces allocations and CPU time. A normalization helper is used only when nested tags are present to maintain functional parity (expanding self-closing tags).
- **XLIFF Marshaling**: Removed the `collectXLIFFUnitTargets` pre-pass. The marshaler now buffers tokens for each `trans-unit` or `unit` and decides whether to update the source or target based on the buffered content. This makes the operation O(N) instead of O(2N).
- **Line Numbering**: Replaced manual byte scanning with `strings.Count` for a significant speedup in error reporting paths.

Performance metrics for 1000 units:
- Parse: 19.7ms/op -> 6.1ms/op
- Marshal: 13ms/op -> 8.6ms/op

---
*PR created automatically by Jules for task [17902570217331589079](https://jules.google.com/task/17902570217331589079) started by @cungminh2710*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -106,3 +106,7 @@
 ## 2026-07-30 - Caching strings.Replacer for static string escaping
 **Learning:** `strings.NewReplacer` performs pre-computation (building a trie) on initialization. Rebuilding it inside a function called frequently (like `encodeAppleStringsQuoted`) introduces significant overhead. Moving it to a package-level variable allows the trie to be built once and reused.
 **Action:** Moved `strings.NewReplacer` to a package-level variable in `internal/i18n/translationfileparser/strings_parser.go`, resulting in a ~9x speedup for string escaping.
+
+## 2026-05-31 - Safe XLIFF token buffering and raw slicing
+**Learning:** Go's `xml.Decoder` reuses internal buffers for tokens (like attributes). Storing tokens in a slice for later processing (e.g., to eliminate a second pass in `MarshalXLIFF`) requires deep cloning via a `cloneXMLToken` helper to avoid data corruption. Additionally, `xml.Encoder` by default expands self-closing tags (e.g., `<ph/>` to `<ph></ph>`), so raw slicing in `Parse` requires a normalization step for elements with nested markup to maintain functional parity with previous behavior.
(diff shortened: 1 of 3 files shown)


cungminh2710:

---

### Issue 1 of 2
internal/i18n/translationfileparser/xliff_parser.go:175-182
**Silent translation corruption for units keyed `_source_locale` / `_target_locale`**

`marshalValues["_source_locale"] = sourceLocale` overwrites any entry in `values` whose unit ID matches `_source_locale` (same for `_target_locale`). Inside `marshalXLIFFUnit`, `replacement, hasReplacement := values[unitKey]` reads from this merged map, so a unit whose `id` attribute is `_source_locale` would have its translation silently replaced by the locale tag string (e.g. `"en-US"`) instead of the intended text. The injected keys are also unused: `rewriteXLIFFLocaleAttrs` only modifies `<file>` and `<xliff>` elements, which never appear as descendants of `<trans-unit>` or `<unit>`, making the call inside `marshalXLIFFUnit` always a no-op. The locale strings should be passed as explicit parameters to `marshalXLIFFUnit` instead of through the shared map.

### Issue 2 of 2
internal/i18n/translationfileparser/xliff_parser.go:127-137
`enc.EncodeToken` and `enc.Flush` errors are silently swallowed with `_ =`. If encoding fails (e.g. unbalanced nesting due to malformed content), the function returns whatever partial bytes have been written to `out` with no indication of failure, which can silently corrupt the stored translation value.

```suggestion
	enc := xml.NewEncoder(&out)
	dec := xml.NewDecoder(bytes.NewReader(val))
	for {
		tok, err := dec.Token()
		if err != nil {
			break
		}
		if err := enc.EncodeToken(tok); err != nil {
			return val
		}
	}
	if err := enc.Flush(); err != nil {
		return val
	}
	return out.Bytes()
```

PR #874, opened 2026-06-03
⚡ Bolt: optimize JSON parser performance

cungminh2710:
This optimization improves the performance of the JSON translation file parser by eliminating redundant iterations and unnecessary sorting. It specifically targets the `JSONParser` used for both standard nested JSON and FormatJS catalogs.

Key changes:
1.  **Single-pass FormatJS parsing**: Previously, FormatJS files were checked for their schema, then messages were extracted, and finally descriptions were extracted in separate passes. These are now combined into a single loop in the `parseFormatJS` helper.
2.  **Removal of unnecessary sorting**: Both `flattenJSON` and the previous FormatJS parsing logic sorted keys before iteration. Since the output is a `map[string]string`, this $O(N \log N)$ operation was wasteful.
3.  **Map capacity hinting**: Provided hints to `make(map[string]string, n)` where the size is known or estimable, reducing allocation overhead during map growth.
4.  **Improved robustness**: Replaced direct type assertions (e.g., `.(string)`) with safe comma-ok assertions to prevent panics on malformed input, returning descriptive errors instead.

The changes were verified with existing unit tests and a new benchmark suite (run during development), showing a measurable performance gain of ~16-17%.

---
*PR created automatically by Jules for task [5452165469211699621](https://jules.google.com/task/5452165469211699621) started by @cungminh2710*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -120,3 +120,7 @@
 ## 2026-08-10 - Optimizing JSONC comment parsing and stack management
 **Learning:** In recursive or stateful parsing (like JSONC comment extraction), repeated use of `bytes.Split` and `strings.Join` for path management (e.g., `stackPrefix`) leads to significant allocation overhead. Replacing `bytes.Split` with manual `bytes.IndexByte` iteration and managing the stack prefix incrementally (concatenating on push, slicing on pop) provides measurable efficiency gains.
 **Action:** Optimized `parseJSONCKeyComments` in `internal/i18n/translationfileparser/jsonc_parser.go`, achieving ~7.4% faster parsing and ~8% fewer allocations.
+
+## 2026-08-15 - Optimizing JSON and FormatJS parsing
+**Learning:** Sorting keys before iterating over a map to populate another map (like in JSON flattening or FormatJS extraction) adds O(N log N) overhead and extra allocations for no benefit, as Go maps are unordered. Additionally, combining multiple validation and extraction passes into a single loop significantly reduces CPU time for specialized formats.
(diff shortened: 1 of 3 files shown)


cungminh2710:

---

### Issue 1 of 2
internal/i18n/translationfileparser/json_parser.go:47-83
**Duplicated FormatJS detection logic**

`parseFormatJS` inlines the same schema check that `IsStrictFormatJSRoot` performs: value must be `map[string]any`, `defaultMessage` must be present and a string. Previously `parseStrictFormatJSMessages` delegated to the private `isStrictFormatJSRoot`, so changes to the schema propagated automatically to both the parse and marshal paths. Now `parseFormatJS` (used by `ParseWithContext`) and `IsStrictFormatJSRoot` (used by `MarshalJSON`) are independent — a future addition such as a required `id` field in the schema would need to be applied to both functions, or the two code paths will silently disagree on what constitutes a valid FormatJS catalog.

### Issue 2 of 2
internal/i18n/translationfileparser/json_parser.go:52
**Eager allocation discarded for non-FormatJS payloads**

`out` is allocated at capacity `len(payload)` before any key has been validated. For a non-FormatJS file (any file where the first visited key lacks a `defaultMessage` field or is not an object), this allocation is immediately discarded and `ParseWithContext` then allocates a second map on line 39. The old code avoided this by passing a single `out` map into `parseStrictFormatJSMessages` and reusing it for `flattenJSON`. For large regular JSON files this adds one extra heap allocation per parse call that the benchmark may not have captured if it only exercised FormatJS inputs.

PR #876, opened 2026-06-03
fix(crowdin): improve machine translation engines API v2 parity

cungminh2710:
This PR improves the Machine Translation Engines (MTE) API v2 parity in the Crowdin fork. 

Key changes:
- **Credential Flexibility:** Changed `MachineTranslation.Credentials` to `map[string]any` and `MTAddRequest.Credentials` to `any`. This allows the SDK to support various MT engines (Google, DeepL, Microsoft, etc.) which have different credential structures, moving away from a hardcoded struct that only supported Crowdin's own NMT.
- **API Parity:** Added missing standard fields `CreatedAt` and `UpdatedAt` to the `MachineTranslation` model and included `Pagination` in the `MachineTranslationsListResponse`.
- **Test Fixes:** Fixed and unskipped `TestMachineTranslationEnginesService_AddMT` by correcting mock JSON data (changing `isEnabled` from string to boolean) and aligning expectations with the new flexible credential type.

Docs: https://developer.crowdin.com/api/v2/#tag/Machine-Translation-Engines

---
*PR created automatically by Jules for task [12819748868877514136](https://jules.google.com/task/12819748868877514136) started by @cungminh2710*


cungminh2710:

---

### Issue 1 of 1
third_party/crowdin-api-client-go/crowdin/model/machine_translation_engines.go:149-151
Typed nil bypasses the credentials validation. Now that `Credentials` is typed as `any`, a caller who assigns a typed-nil pointer (e.g. `var c *MTECredentials = nil; req.Credentials = c`) will have `r.Credentials == nil` evaluate to `false` — the interface box holds a non-nil type descriptor even though the underlying value is nil. The validation silently passes and the JSON-marshaled body will contain `"credentials":null`, which the Crowdin API will likely reject at runtime rather than at the SDK boundary.

```suggestion
	if r.Credentials == nil || (reflect.ValueOf(r.Credentials).Kind() == reflect.Ptr && reflect.ValueOf(r.Credentials).IsNil()) {
		return errors.New("credentials are required")
	}
```

PR #877, opened 2026-06-03
🧪 Scout: support triple mustache placeholders

cungminh2710:
💡 What: Added support and unit tests for triple mustache placeholders in the ICU invariant parser.
🎯 Why: Mustache-based systems often use `{{{key}}}` for unescaped content. Supporting this syntax prevents validation failures in translations that use it.
🧩 Scope: `internal/i18n/icuparser/invariant.go`, `internal/i18n/icuparser/invariant_test.go`
✅ Verification: Ran `go test -v ./internal/i18n/icuparser/...`. All tests passed.
🛡️ Confidence: Prevents false-positive validation errors for projects using triple-mustache placeholders.

---
*PR created automatically by Jules for task [3470624262101809011](https://jules.google.com/task/3470624262101809011) started by @cungminh2710*

### DIFF
--- a/.jules/scout.md
+++ b/.jules/scout.md
@@ -43,3 +43,11 @@
 ## 2025-06-12 - [PO msgid Significance of Whitespace]
 **Learning:** In gettext/PO files, `msgid` keys are the source of truth for translation lookups, and leading/trailing whitespace is significant. Over-eagerly trimming spaces from these keys during parsing causes lookup failures in downstream systems.
 **Action:** Always preserve the exact literal string for `msgid` keys, except for the header entry (`msgid ""`) which is standardly skipped in message maps.
+
+## 2026-06-03 - [Triple Mustache Placeholder Normalization]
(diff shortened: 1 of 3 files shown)


cungminh2710:

---

### Issue 1 of 2
.jules/scout.md:47-53
**Duplicate and malformed scout.md entry**

The same learning has been appended twice (lines 47–49 and 51–53). Additionally, the first copy has corrupted text — "uses a fallback" is missing the subject (`ParseInvariant`) before each gap — suggesting it was generated with dropped words. One clean entry should replace both.

### Issue 2 of 2
internal/i18n/icuparser/invariant_test.go:124-128
**Missing edge-case tests for triple mustache**

The new test covers the happy path but leaves two important cases untested: (1) an invalid triple-mustache like `{{{not a valid id}}}` (spaces in name) — the analogous double-mustache test already exists at line 140, and the triple case should behave the same way; (2) mixed double and triple placeholders in a single string, e.g. `"{{name}} and {{{title}}}"`, to confirm neither branch interferes with the other. Without these, a regression that changes how `isPlaceholderName` failure falls through for triple-braces would be invisible.

PR #885, opened 2026-06-04
feat(crowdin): add user account linking

cungminh2710:
## What changed

- Added per-user Crowdin OAuth linking alongside the organization-level Crowdin integration.
- Show a navbar CTA when Crowdin is connected for the workspace and the current user has not linked Crowdin.
- Use the linked Crowdin identity for My Jobs matching and user-attributed Crowdin task description edits.

## How to test

- `vp run db:migrate`
- `vp check --fix`
- `vp test src/api/routes/external-tms-provider-credential/external-tms-provider-credential.test.ts`
- `vp test`

## Checklist

### DIFF
--- a/apps/hyperlocalise-web/drizzle/0042_worried_pete_wisdom.sql
+++ b/apps/hyperlocalise-web/drizzle/0042_worried_pete_wisdom.sql
@@ -1,0 +1,45 @@
+CREATE TABLE "crowdin_user_connections" (
+	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+	"organization_id" uuid NOT NULL,
+	"user_id" uuid NOT NULL,
+	"provider_credential_id" uuid NOT NULL,
+	"crowdin_user_id" integer NOT NULL,
+	"username" text NOT NULL,
+	"email" text,
+	"full_name" text,
+	"oauth_expires_at" timestamp with time zone,
+	"encryption_algorithm" text NOT NULL,
+	"ciphertext" text NOT NULL,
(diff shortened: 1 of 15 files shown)


cungminh2710:

---

### Issue 1 of 2
apps/hyperlocalise-web/src/api/routes/external-tms-provider-credential/external-tms-provider-credential.route.ts:505-521
**Unhandled unique constraint violation crashes the OAuth callback**

`upsertCrowdinUserConnection` uses `onConflictDoUpdate` that targets only `(organizationId, userId)`. There is a second unique index on `(organizationId, crowdinUserId)`. If a second workspace member tries to link the same Crowdin account (same Crowdin user ID), the INSERT will conflict on `crowdin_user_connections_org_crowdin_user_key`, which is not handled by the `onConflictDoUpdate` clause, so Postgres throws a raw unique-constraint error. Because this call is not wrapped in a try/catch, the exception propagates unhandled and the user receives a 500 HTTP response instead of a descriptive redirect.

### Issue 2 of 2
apps/hyperlocalise-web/src/components/app-shell/app-shell.tsx:24-40
**Two extra DB queries on every page render for Crowdin-connected orgs**

`AppShell` is a Next.js server component, so it executes on every navigation. When the active credential is a Crowdin OAuth integration (which is the common case after setup), this adds two sequential DB queries — `getActiveOrganizationExternalTmsProviderCredentialRow` and `getCrowdinUserConnectionSummary` — to every page load for every member of that workspace. Consider caching the result or bundling the user-connection lookup into the existing credential query so the navbar overhead does not compound across high-traffic orgs.

PR #898, opened 2026-06-05
⚡ Bolt: optimize line splitting and CRLF normalization

cungminh2710:
💡 What: Optimized Markdown, MDX, and YAML parsers by replacing `strings.SplitAfter` and `bytes.Split` with manual `IndexByte` loops. Implemented fast-path checks for CRLF normalization and replaced expensive `[]byte` -> `string` -> `[]byte` conversions with `bytes.ReplaceAll`. Added an ASCII fast-path to `isTranslatableChunk`.

🎯 Why: Large translation files were causing significant memory pressure due to large string/byte slice allocations during line splitting. Unconditional CRLF replacement was causing redundant allocations for files that were already normalized.

📊 Impact:
- `collectFrontmatterCandidates` allocations reduced by ~15% (2.0MB -> 1.7MB for 1MB test file).
- `detectYAMLIndent` allocations reduced to zero (483KB -> 0B).
- `MDXParse` speed improved by ~3% (~77ms -> ~75ms).

🔬 Measurement: Verified with `go test ./internal/i18n/...` and benchmarks in `internal/i18n/translationfileparser/bolt_benchmark_test.go` (temporary benchmark file).

---
*PR created automatically by Jules for task [15702951437490496207](https://jules.google.com/task/15702951437490496207) started by @cungminh2710*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -128,3 +128,7 @@
 ## 2026-08-20 - Optimizing CSV parsing and marshaling via streaming
 **Learning:** Loading entire files into memory using `csv.ReadAll` is a major bottleneck for large translation files. A streaming approach using `csv.Reader.Read` and `csv.Writer.Write` allows processing files row-by-row, significantly reducing peak memory usage.
 **Action:** Refactored `CSVParser.Parse` and `MarshalCSV` to use streaming I/O, resulting in ~32% fewer allocations for parsing and ~38% fewer for marshaling, while improving marshaling speed by ~32%.
+
+## 2026-08-25 - Optimizing Markdown and YAML line processing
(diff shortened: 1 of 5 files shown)


cungminh2710:

---

### Issue 1 of 2
internal/i18n/translationfileparser/markdown_mdx_parser.go:105-119
The `if line == ""` check is now unreachable dead code. In the `IndexByte` loop, when `len(text) > 0` the `line` variable is always assigned at least one character — either the full remaining text (no newline) or `text[:idx+1]` which contains at least the `
` itself. The guard was meaningful with `SplitAfter` because it produced a trailing empty string for inputs ending in `
`, but that never happens here. The same dead guard appears in `emitMarkdownText` below.

```suggestion
	// BOLT OPTIMIZATION: Avoid strings.SplitAfter(text, "\n") to reduce allocations.
	for len(text) > 0 {
		var line string
		idx := strings.IndexByte(text, '\n')
		if idx < 0 {
			line = text
			text = ""
		} else {
			line = text[:idx+1]
			text = text[idx+1:]
		}
		emitFrontmatterLineParts(line, s.doc, func(part markdownPart) {
```

### Issue 2 of 2
internal/i18n/translationfileparser/markdown_mdx_parser.go:133-147
Same dead `if line == ""` guard in `emitMarkdownText` — `line` is always non-empty inside this loop.

```suggestion
	// BOLT OPTIMIZATION: Avoid strings.SplitAfter(text, "\n") to reduce allocations.
	for len(text) > 0 {
		var line string
		idx := strings.IndexByte(text, '\n')
		if idx < 0 {
			line = text
			text = ""
		} else {
			line = text[:idx+1]
			text = text[idx+1:]
		}
		trimmed := strings.TrimSpace(line)
```

PR #899, opened 2026-06-05
feat(canva-app): add text translation template app

cungminh2710:
## Summary

Adds a placeholder Canva Apps SDK project at `apps/canva-app`, based on the [official text translation example](https://www.canva.dev/docs/apps/examples/text-translation/).

The app demonstrates bulk text editing on the current page via `editContent`, with two modes:

- **Translate with formatting** — preserves inline formatting by translating text regions individually
- **Translate without formatting** — replaces plain text in each richtext element

Translation is simulated with lorem ipsum placeholder text (not a real translation service).

## Tooling

- **License:** Business Source License 1.1 (`LICENSE`)
- **Toolchain:** Vite+ (`vp`) with oxlint, oxfmt, and Vitest
- **Canva bundling:** webpack via `vp run start` and `vp run bundle` (required by the Apps SDK)

## How to run

```bash
cd apps/canva-app
vp install
vp run start
```

Set the Development URL to `http://localhost:8080` in the Canva Developer Portal to preview.

## Testing

- `vp check --fix` — passes
- `vp test` — passes (2 tests)

<div><a href="https://cursor.com/agents/bc-8cb9d069-608f-4de9-a1d1-ee275cd9f31c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-8cb9d069-608f-4de9-a1d1-ee275cd9f31c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


cungminh2710:

---

### Issue 1 of 3
apps/canva-app/src/intents/design_editor/app.tsx:15-38
**No error handling — buttons permanently stuck on failure**

Both `translateWithoutFormatting` and `translateWithFormatting` call `setInProgressTask(undefined)` only at the very end of the `async` function. If `editContent` or `getTranslation` throws, that final call is never reached, leaving `inProgressTask` set forever. The buttons will remain disabled and one will stay in the loading state for the rest of the component's lifetime, with no way for the user to retry.

The same issue exists in `translateWithFormatting` (lines 41–78). Wrap the body in a `try/finally` block so `setInProgressTask(undefined)` runs unconditionally.

### Issue 2 of 3
apps/canva-app/pnpm-workspace.yaml:1-4
**Placeholder values left in `allowBuilds`**

Both `core-js-pure` and `esbuild` are set to the literal string `set this to true or false` instead of a boolean `true` or `false`. pnpm's `allowBuilds` map expects boolean values; a non-boolean value may be coerced to `true` (enabling build scripts you may not intend to run) or cause a validation error on `vp install`. This needs to be resolved before the workspace is usable.

### Issue 3 of 3
apps/canva-app/scripts/ssl/ssl.ts:42-50
Non-standard RSA modulus length. Common, interoperable choices are 2048 (standard minimum) or 4096 (high security). `2096` is almost certainly a typo of `2048` — it's not a power of two and is not recognized as a standard key size by most crypto tooling or browsers, which could cause the self-signed certificate to be rejected in some environments.

```suggestion
const generateRsaKeys = async (): Promise<{
  publicKey: string;
  privateKey: string;
}> =>
  new Promise((resolve, reject) => {
    crypto.generateKeyPair(
      "rsa",
      {
        modulusLength: 2048,
```

cungminh2710:

---

### Issue 1 of 3
apps/canva-app/scripts/ssl/ssl.ts:6
SSL certificate directory resolves two levels above the app root, placing the generated files at `<monorepo root>/.ssl` instead of `apps/canva-app/.ssl`. The project-level `.gitignore` only covers `.ssl` relative to `apps/canva-app/`, so the generated private key at the monorepo root is NOT gitignored by that file — and the root `.gitignore` has no `.ssl` entry either. Running `git add .` from the repo root after `vp run start --use-https` would silently stage the private key file.

```suggestion
const SSL_CERT_DIR = path.resolve(process.cwd(), ".ssl");
```

### Issue 2 of 3
apps/canva-app/src/intents/design_editor/app.tsx:61-76
**`endOfRegion` accumulation breaks when text regions don't cover the full plaintext**

The loop subtracts `region.text.length` from `endOfRegion` (which starts at `range.readPlaintext().length`) to reconstruct each region's start offset. This arithmetic only holds when every character in the plaintext is contained in a contiguous, back-to-back series of text regions. If the SDK yields regions with gaps between them (e.g., an unstyled space or punctuation not belonging to any formatted run), every `replaceText` call after the first gap will target a wrong index, silently mis-applying translations to incorrect character positions.

### Issue 3 of 3
apps/canva-app/src/intents/design_editor/lorem_generator.ts:15-17
**Unsafe array access produces `undefined` at type boundary**

`loremWords[index]` has a TypeScript type of `string | undefined` (since TypeScript does not narrow fixed-length plain arrays), yet the function's return type is `string[]` and the caller uses the values directly as `translatedText`. The modulo ensures the index stays in range at runtime, but adding a non-null assertion (`loremWords[index]!`) or a fallback (`?? ""`) makes the intent explicit and avoids silent widening if the compiler setting changes.

PR #915, opened 2026-06-06
fix(crowdin): improve StringComment model parity for hasPlurals and isIcu

cungminh2710:
Improved the parity of the Crowdin Go SDK with the official API v2 by updating the `String` model within the `StringComments` service to include missing `hasPlurals` and `isIcu` fields. Also corrected an erroneous documentation URL in the `Get` method of the `StringComments` service. verified with unit tests and pre-commit checks.

---
*PR created automatically by Jules for task [13366744648752931074](https://jules.google.com/task/13366744648752931074) started by @cungminh2710*


cungminh2710:

---

### Issue 1 of 1
third_party/crowdin-api-client-go/crowdin/string_comments_test.go:334-336
**Test values don't validate JSON deserialization of the new fields**

Both `HasPlurals` and `IsIcu` are set to `false` in the fixture, but `false` is also the zero value for `bool` in Go. If the JSON struct tags were wrong or entirely missing, the fields would still silently default to `false` and the test would pass unchanged. A quick variant that asserts `true` roundtrips correctly (or at least flipping the fixture JSON to `true` for one of the fields) would prove the tags actually work.

PR #916, opened 2026-06-06
🧪 Scout: Support array indices in ICU placeholders

cungminh2710:
💡 What: Added support for array indices (e.g., `{items[0]}`) in ICU placeholders and added a regression test.
🎯 Why: Flattened JSON data often uses indexed keys, and the ICU parser was failing to recognize these as valid placeholders, leading to incomplete invariant extraction.
🧩 Scope: `internal/i18n/icuparser/invariant.go` and `internal/i18n/icuparser/invariant_test.go`.
✅ Verification: Ran `go test ./internal/i18n/icuparser/...` and verified all tests pass, including the new `TestPlaceholderWithArrayIndices`.
🛡️ Confidence: High. The change is narrow and specifically targets the validation logic for placeholder names.

---
*PR created automatically by Jules for task [7666540753325459650](https://jules.google.com/task/7666540753325459650) started by @cungminh2710*

### DIFF
--- a/internal/i18n/icuparser/invariant.go
+++ b/internal/i18n/icuparser/invariant.go
@@ -382,7 +382,7 @@ func isASCIIPlaceholderFirst(b byte) bool {
 }
 
 func isASCIIPlaceholderSubsequent(b byte) bool {
-	return (b >= 'a' && b <= 'z') || (b >= 'A' && b <= 'Z') || (b >= '0' && b <= '9') || b == '_' || b == '.' || b == '-' || b == '$'
+	return (b >= 'a' && b <= 'z') || (b >= 'A' && b <= 'Z') || (b >= '0' && b <= '9') || b == '_' || b == '.' || b == '-' || b == '$' || b == '[' || b == ']'
 }
 
 func isASCIIDigit(b byte) bool {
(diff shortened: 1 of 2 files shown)


cungminh2710:

---

### Issue 1 of 2
internal/i18n/icuparser/invariant.go:384-386
**No bracket structural validation**

Adding `[` and `]` as accepted subsequent characters allows any arrangement of those characters after the first byte — including `items[`, `items]`, `items][`, `items[0][`, and `items[[0]]`. The validation has no structural check that brackets are balanced or that only digits appear inside them. Because `readIdentifierLike` stops at `,` before `}`, a message like `{data[0, plural, …}` would parse the argument name as `data[0` (open bracket, no closer), and that string would now pass `isPlaceholderName` and be silently collected as a placeholder named `data[0`. A narrower predicate — for example, enforcing that `[` must be followed by one-or-more ASCII digits and then `]` — would admit only valid array index notation while rejecting the malformed forms.

### Issue 2 of 2
internal/i18n/icuparser/invariant_test.go:438-479
**Edge-case and moustache paths not covered**

The new test only exercises the ICU-style path (`{items[0]}`). Two additional scenarios would provide more complete coverage: (1) moustache-style input (`{{items[0]}}`) routed through `normalizeMustachePlaceholders`, since that codepath also calls `isPlaceholderName`; and (2) malformed bracket input such as `{items[}` or `{data[0, plural, one {# item} other {# items}}` to document the current acceptance behaviour and catch any future tightening of the validator.

PR #918, opened 2026-06-06
feat(slack): track image assets for multi-turn localization

cungminh2710:
## Summary

Slack image localization now persists source and translated images across thread turns, so users can clarify a target locale or request another localization without reuploading.

## Changes

- Store uploaded Slack images as `stored_files` with role `source` and metadata `imageLocalizationSource: true`
- Store generated localized images as `stored_files` with role `output`, linked to the source via `sourceFileId` and `targetLocale`
- Track thread state in `SlackBotThreadState`:
  - `pendingSlackImageTask` when an image is uploaded without a target locale
  - `imageSourceAssets` with localized output history per source image
- Handle text-only follow-ups via `handleSlackImageFollowUp` (e.g. "Localize to French" or "Now do Japanese")
- Update missing-locale prompt to ask for a reply instead of reuploading
- Add `getStoredFileContent` helper for reloading stored image bytes

## Review fixes

- Pass attachment index into `storeSlackImageSource` so batch nameless uploads get unique fallback filenames
- Use a local `threadState` variable in `localizeSlackImageSources` instead of mutating the input object
- Remove expiring `slackAttachmentUrl` from persisted source metadata
- Call `wrapThreadPost` only when image follow-up will localize (via `beforeLocalize` callback)

## User flows enabled

1. Upload image → bot asks for language → user replies "French" → localization runs from stored source
2. Upload + localize to French → user replies "Now localize to Japanese" → new output from same source
3. Existing single-turn upload + localize behavior unchanged

## Testing

- `vp test src/lib/agents/slack/bot.test.ts` (41 tests)
- `vp check --fix`

<div><a href="https://cursor.com/agents/bc-cd62a301-6f85-476c-8bfd-a46c9360cc20"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-cd62a301-6f85-476c-8bfd-a46c9360cc20"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/hyperlocalise-web/src/lib/agents/slack/bot.test.ts
+++ b/apps/hyperlocalise-web/src/lib/agents/slack/bot.test.ts
@@ -11,7 +11,7 @@ import {
   handleSubscribedMessage,
   wrapThreadPost,
 } from "./bot";
-import { getSlackRepositoryContextKey } from "./repository-session";
+import { getSlackRepositoryContextKey, type SlackBotThreadState } from "./repository-session";
 
 function createMockClassification(
   overrides: Partial<ConversationClassification> = {},
@@ -136,12 +136,24 @@ vi.mock("@/lib/agent-runtime/workspaces/repository-sandbox", () => ({
 }));
 
 vi.mock("@/lib/file-storage/records", () => ({
-  createStoredFile: vi.fn(async (input: { filename: string; contentType: string }) => ({
-    id: "file_123",
-    filename: input.filename,
-    contentType: input.contentType,
-    downloadUrl: null,
-    storageUrl: "https://files.example/file_123",
+  createStoredFile: vi.fn(
+    async (input: { filename: string; contentType: string; role?: string }) => ({
+      id: input.role === "output" ? "file_output_123" : "file_123",
+      filename: input.filename,
+      contentType: input.contentType,
+      downloadUrl: null,
+      storageUrl: "https://files.example/file_123",
+      storageKey: "organizations/org/files/file_123/banner.png",
+    }),
+  ),
+  getStoredFileContent: vi.fn(async () => ({
+    file: {
+      id: "file_123",
+      filename: "banner.png",
+      contentType: "image/png",
+      storageKey: "organizations/org/files/file_123/banner.png",
+    },
+    content: Buffer.from("source-image"),
   })),
 }));
 
@@ -206,7 +218,7 @@ vi.mock("@/lib/database", () => {
 
 import { generateText } from "ai";
 import { findSlackConnector, lookupMembership } from "@/lib/agents/slack/helpers";
-import { createStoredFile } from "@/lib/file-storage/records";
+import { createStoredFile, getStoredFileContent } from "@/lib/file-storage/records";
 import { regenerateImageFromAttachment } from "@/lib/agents/image-generation";
 import {
   addInteractionMessage,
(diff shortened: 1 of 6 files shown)


cungminh2710:
@cursoragent Fix the following 4 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 4
apps/hyperlocalise-web/src/lib/agents/slack/image-session.ts:47
**Hardcoded index produces duplicate fallback filenames for batch uploads**

`attachmentFilename` uses the index only as a fallback when `attachment.name` is falsy — it is always called with `0` here, so every nameless attachment in a same-message `Promise.all` batch gets stored as `slack-image-1.png`. The duplicated filename is written into `imageSourceAssets`, so follow-up reply messages will show ambiguous names (e.g., "banner-fr.webp" referring to different source images). The `storeSlackImageSource` signature needs an `index` parameter passed through from the `.map` callback in `handleSlackImageAttachments`.

### Issue 2 of 4
apps/hyperlocalise-web/src/lib/agents/slack/image-attachments.ts:271-286
`input.threadState` is mutated directly on the shared `input` object. A local variable better expresses the intent and avoids the implicit side-effect on the caller's data structure.

```suggestion
  let localizedCount = 0;
  let threadState = input.threadState;

  for (const source of input.sources) {
    try {
      await localizeSlackImageSource({
        thread: input.thread,
        message: input.message,
        source,
        targetLocale: input.targetLocale,
        instructions: input.instructions,
        threadState,
        storage: input.storage,
        beforePostGeneratedImage: input.beforePostGeneratedImage,
      });
      localizedCount += 1;
      threadState = (await input.thread.state) as SlackBotThreadState | null;
```

### Issue 3 of 4
apps/hyperlocalise-web/src/lib/agents/slack/image-session.ts:63-65
**Expiring Slack URL stored permanently in file metadata**

`slackAttachmentUrl` captures the raw Slack attachment URL, which Slack revokes after a short window (typically a few hours). The actual file bytes are already persisted correctly in the storage adapter, so this field has no functional use — but any future code or tooling that reads `metadata.slackAttachmentUrl` will silently receive a dead link. Consider omitting the field, or documenting that it is informational-only and may have expired.

### Issue 4 of 4
apps/hyperlocalise-web/src/lib/agents/slack/bot.ts:506-521
**`wrapThreadPost` called before knowing if the follow-up will be handled**

In the `else if (threadHasStoredSlackImages(threadState))` branch, `wrapThreadPost` is called unconditionally. If `handleSlackImageFollowUp` returns `{ handled: false }` (the message contains no target locale), execution falls through to the main agent flow, which calls `wrapThreadPost` a second time at the end of `processSlackMessage`. The pre-existing image-attachments path shares this pattern, so it appears idempotent in practice — but moving `wrapThreadPost` to inside the `if (followUpResult.handled)` guard (or after an explicit early-return check) would make the intent explicit and match how the function is used in the non-follow-up paths.

PR #922, opened 2026-06-07
feat(cat): add next-gen CAT workspace UI with Storybook

cungminh2710:
## Summary

Adds a next-generation CAT (computer-assisted translation) workspace UI as a reusable component library under `src/components/cat/`, with Storybook stories for design review and iteration.

The layout matches the three-column spec from the design reference:

- **Left — Queue:** segment list with status, progress, and sequential review entry point
- **Center — Editor:** source/target panes, AI suggestion, format & QA checks, review actions, and suggestions/history/glossary tabs
- **Right — Intelligence:** context, glossary terms, QA risks, GitHub evidence, and related strings

## Architecture

- **`CatWorkspaceView`** — presentational shell; all behavior is injected via `CatWorkspaceDependencies`
- **`CatWorkspaceContainer`** — optional stateful wrapper for Storybook and local prototyping (segment navigation, target edits, approve/skip, async format validation)
- **DI interfaces** — `navigation`, `editing`, `review`, `toolbar`, and `services.validateFormat` for later API/TMS integration without changing UI components

## Storybook

Stories live at **CAT/Workspace**:

- `Default` — full workspace with fixture data and action mocks
- `EmptyQueue` — empty state
- `InteractiveReview` — approve flow interaction test

Run with:

```bash
cd apps/hyperlocalise-web && vp run storybook
```

## Validation

- `vp check --fix` passes
- `vp run build-storybook` passes

<div><a href="https://cursor.com/agents/bc-c344d486-26fc-4817-ab50-b5ccf3b08e14"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-c344d486-26fc-4817-ab50-b5ccf3b08e14"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


cungminh2710:

---

### Issue 1 of 1
apps/hyperlocalise-web/src/components/cat/cat-target-editor.tsx:79-97
The offset counter in `descendants` only accumulates raw text-node lengths, but `textBetween(..., "
", "
")` inserts a `"
"` separator between every pair of block nodes. For a document with two paragraphs the second paragraph's text starts at `firstLen + 1` in the analyzed string, yet the `textRanges` entry records `offsetStart = firstLen`. Any token whose `start`/`end` falls inside the second (or later) paragraph will be shifted one position too early per preceding paragraph, placing the decoration one character off per paragraph boundary.

```suggestion
              const textRanges: Array<{
                offsetStart: number;
                offsetEnd: number;
                posStart: number;
              }> = [];
              let offset = 0;
              let prevNodeEnd = -1;

              state.doc.descendants((node, pos) => {
                if (!node.isText || !node.text) {
                  return;
                }

                // Account for the "\n" block separator that textBetween inserts
                // between consecutive text nodes that belong to different blocks.
                if (prevNodeEnd !== -1 && pos !== prevNodeEnd) {
                  offset += 1; // one separator character per block boundary
                }

                textRanges.push({
                  offsetStart: offset,
                  offsetEnd: offset + node.text.length,
                  posStart: pos,
                });
                offset += node.text.length;
                prevNodeEnd = pos + node.text.length;
              });
```

cungminh2710:
@cursoragent Fix the following 4 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 4
apps/hyperlocalise-web/src/components/cat/cat-message-format.ts:239-258
**ICU token signature too strict — valid locale-appropriate plural forms will be flagged as errors**

`tokenSignature` for `icu` tokens includes the sorted option keys (`one|other`), so a Vietnamese or Russian translation that correctly adds locale-specific plural categories (e.g. `few`, `many`) would produce a different signature from the English source, causing `findMissingTokens` to report the source block as "missing" from the target and emit an `icu-mismatch` error even though the translation is valid.

For example, a source `{count, plural, one {# item} other {# items}}` produces signature `icu:count:plural:one|other`. A correct Russian translation `{count, plural, one {# элемент} few {# элемента} many {# элементов} other {# элемента}}` produces `icu:count:plural:few|many|one|other`. These don't match, and the source block is flagged as missing. ICU structure checks should verify that the same argument name and type exist in the target, not that the option set is identical.

### Issue 2 of 4
apps/hyperlocalise-web/src/components/cat/cat-workspace-container.tsx:167-174
`onUseAiSuggestion` reads `aiSuggestion` from the `useMemo` closure (`state.intelligence.aiSuggestion`) rather than `stateRef.current`. Using `stateRef` consistently for all state reads inside callbacks would eliminate any stale-closure risk if `aiSuggestion` changes between renders without triggering the memo.

```suggestion
      onUseAiSuggestion: (segmentId: string) => {
        const aiSuggestion = stateRef.current.intelligence.aiSuggestion;
        if (!aiSuggestion) {
          return;
        }
        editing.onTargetChange(segmentId, aiSuggestion);
        onUseAiSuggestion?.(segmentId);
      },
```

### Issue 3 of 4
apps/hyperlocalise-web/src/components/ui/kbd.tsx:16
`KbdGroup` declares its props as `React.ComponentProps<"div">`, which types the ref as `HTMLDivElement`. The element actually rendered is `<kbd>`, so the ref type is wrong and any `div`-specific types that differ from `kbd` will silently pass TypeScript checks. The props type should match the rendered element.

```suggestion
function KbdGroup({ className, ...props }: React.ComponentProps<"kbd">) {
```

### Issue 4 of 4
apps/hyperlocalise-web/src/components/cat/cat-workspace-container.tsx:90-113
**`isBusy` never set — action buttons are always active during async validation**

`runSegmentChecks` fires an async `Promise.all` on every target change but the container never passes `isBusy={true}` to `CatWorkspaceView`. As a result the Approve, Previous, and Next buttons remain fully enabled while validation is in flight. A user can approve a segment before the format-check results resolve; the sequence-number guard only prevents stale `formatChecks` writes, it does not prevent this approve-before-validate race. Tracking an `isBusy` boolean in local state (set before the `await`, cleared in a `finally` block) and passing it to `CatWorkspaceView` would close the gap.

cungminh2710:

---

### Issue 1 of 1
apps/hyperlocalise-web/src/components/cat/cat-target-editor.tsx:278-282
The extension factory is called inline inside `useEditor`, creating a new `Extension` instance on every render of `CatTargetEditor`. The same PR deliberately hoists `markdownExtensions` to module scope in `markdown-description-editor.tsx` for this exact reason. While TipTap only reads extensions on initial mount (so the editor is not re-initialized), each render still constructs and discards a new extension object plus a new `StarterKit.configure(...)` instance. Hoisting to module scope avoids the waste and keeps the pattern consistent.

```suggestion
const catTargetEditorExtensions = [
  StarterKit.configure({ heading: false }),
  createCatMessageFormatExtension(),
] as unknown as Extensions;

  const editor = useEditor({
    extensions: catTargetEditorExtensions,
```

PR #924, opened 2026-06-07
⚡ Bolt: optimize GenericXMLParser by reducing allocations in hot paths

cungminh2710:
💡 What: Optimized the `GenericXMLParser` by reducing heap allocations and improving algorithmic efficiency in hot paths. Specifically:
- Replaced `strings.TrimSpace(string(token))` with a custom `isAllXMLWhitespace([]byte)` check for `xml.CharData` and `xml.Comment` tokens, eliminating unnecessary string conversions for indentation and whitespace-only text.
- Refactored `genericXMLKeyAttr` from multiple passes over attributes to a single-pass scan with priority (key > id > name) and fast-paths for common lowercase names.
- Added fast-paths to element and attribute helper functions (`isGenericXMLMetadataElement`, `isGenericXMLLocaleAttrName`, etc.) to skip `strings.ToLower` and `strings.TrimSpace` for common lowercase, trimmed inputs.
- Added a capacity hint to the output map in `Parse` based on the number of discovered entries.

🎯 Why: Generic XML parsing of large files with deep trees or many small elements suffers from high allocation pressure due to frequent string conversions and redundant attribute scans.

📊 Impact: Reduces allocations per element in the XML tree. For large documents with significant indentation, this eliminates one string allocation per indentation block. Single-pass attribute scanning improves performance for elements with multiple attributes.

🔬 Measurement: Run benchmarks using `go test -bench BenchmarkGenericXMLParser -benchmem ./internal/i18n/translationfileparser`. Baseline measurements showed ~29k-34k allocations for a 1000-item document; these optimizations specifically target the per-token overhead within that loop.

---
*PR created automatically by Jules for task [15253095386315803341](https://jules.google.com/task/15253095386315803341) started by @cungminh2710*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -113,6 +113,10 @@
 
 ## 2026-06-01 - Optimizing ARB parsing via single-pass and map hinting
 
+## 2026-09-01 - Optimizing GenericXMLParser via single-pass and allocation avoidance
+**Learning:** XML parsing hot paths, especially `xml.CharData` and attribute scanning, can be significant allocation bottlenecks. Converting `[]byte` tokens to `string` just for whitespace checks or multiple passes over attributes for key discovery adds avoidable overhead. Custom byte-level checks (`isAllXMLWhitespace`) and single-pass priority-based scans are much more efficient.
+**Action:** Use `isAllXMLWhitespace([]byte)` instead of `strings.TrimSpace(string(token))` in XML/HTML parsers and refactor attribute lookups into single-pass scans.
+
 ## 2026-08-05 - Optimizing Fluent parsing and marshaling
 **Learning:** High-level string operations like `strings.Split`, `strings.Join`, and `strings.ReplaceAll` in recursive or iterative document processing (like Fluent parsing) accumulate significant allocation overhead. Replacing them with single-pass loops using `strings.Builder` and pre-allocating slices using `strings.Count` for line counting yields substantial performance gains.
 **Action:** Optimized `scanFluentLines`, `encodeFluentValue`, `normalizeFluentValue`, `formatFluentComments`, and `render` in `internal/i18n/translationfileparser/fluent_parser.go`.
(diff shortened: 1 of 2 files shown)


cungminh2710:

---

### Issue 1 of 2
internal/i18n/translationfileparser/generic_xml_parser.go:446-453
`isAllXMLWhitespace` only checks the four ASCII whitespace bytes (`' '`, `'	'`, `'
'`, `'
'`), while the replaced `strings.TrimSpace(string(token)) != ""` uses Go's `unicode.IsSpace`, which also covers characters like U+00A0 (non-breaking space, encoded as `0xC2 0xA0`), U+0085 (NEL), and the full Unicode White_Space set. A CharData or Comment token that contains only U+00A0 bytes would make `isAllXMLWhitespace` return `false`, so `hasNonWhitespace` / `hasNonTextContent` would be set where the old code would not have set them — potentially misclassifying indentation or spacing nodes in XML files produced by certain editors.

```suggestion
func isAllXMLWhitespace(data []byte) bool {
	for _, b := range data {
		if b >= 0x80 {
			// Non-ASCII byte: fall back to rune-aware Unicode whitespace check.
			for _, r := range string(data) {
				if !unicode.IsSpace(r) {
					return false
				}
			}
			return true
		}
		if b != ' ' && b != '\t' && b != '\n' && b != '\r' {
			return false
		}
	}
	return true
}
```

### Issue 2 of 2
internal/i18n/translationfileparser/generic_xml_parser.go:510
**Redundant len-guard on the "key" fast-path**

`local == "key"` already short-circuits before `len(local) == 3 && strings.EqualFold(local, "key")` is evaluated, so the guard saves nothing on the common path and only activates when `local != "key"`. Since `strings.EqualFold` itself rejects strings of different byte-lengths quickly, the `len` guard is harmless but adds visual noise. The same pattern repeats on the `"id"` and `"name"` branches. Consider using a plain `strings.EqualFold` call, or at least aligning the three branches so each reads identically.

PR #938, opened 2026-06-08
🧪 Scout: Fix ICU plural parsing for negative exact-value selectors

cungminh2710:
### 💡 What
This change enables the ICU parser to correctly handle negative exact-value selectors in plural messages, such as `=-1`.

### 🎯 Why
Previously, the parser only expected digits after the `=` prefix, causing valid ICU messages containing negative explicit values to fail parsing.

### 🧩 Scope
- `internal/i18n/icuparser/parse.go`: Updated `readSelector` to support an optional minus sign.
- `internal/i18n/icuparser/parse_test.go`: Added `TestParseASTPluralNegativeSelector`.
- `.jules/scout.md`: Added journal entry for this learning.

### ✅ Verification
- Ran `go test -v ./internal/i18n/icuparser/ -run TestParseASTPluralNegativeSelector` (Passed)
- Ran all tests in `internal/i18n/icuparser/` (Passed)

### 🛡️ Confidence
High. The fix is a straightforward extension of the existing selector scanning logic and is covered by a new regression test. All existing ICU parser tests passed, ensuring no regressions in standard selector or message parsing.

---
*PR created automatically by Jules for task [10233562293591862890](https://jules.google.com/task/10233562293591862890) started by @cungminh2710*

### DIFF
--- a/.jules/scout.md
+++ b/.jules/scout.md
@@ -61,5 +61,5 @@
 **Action:** When testing identifier validation, include "tail" cases where valid segments are followed by invalid characters to ensure the state machine or regex correctly terminates or rejects the input.
 
 ## 2025-06-18 - [ICU Plural Negative Selectors]
-**Learning:** ICU plural exact-value selectors (e.g., `=-1`) can include negative numbers. A naive selector parser that only expects digits after the `=` prefix will fail on these valid selectors.
-**Action:** Ensure plural selector parsing explicitly handles an optional minus sign following the `=` prefix before consuming digits.
+**Learning:** ICU plural exact-value selectors (e.g., `=-1`) can include negative numbers. A naive selector parser that only expects digits after the `=` prefix will fail on these valid selectors. Additionally, when parsing sequences that require at least one element (like digits after a sign), using an explicit starting position marker (`digitStart`) to verify consumption is clearer and more robust than checking the last character's properties.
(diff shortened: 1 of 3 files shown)


cungminh2710:

---

### Issue 1 of 2
internal/i18n/icuparser/parse.go:699-707
The validity check `p.pos > start+1 && p.src[p.pos-1] != '-'` is correct but encodes "at least one digit was consumed" indirectly through the last-character sentinel. An explicit `digitStart` variable makes the invariant self-evident and removes the need to reason about the `p.src[p.pos-1]` index in the edge case where no digits follow the minus.

```suggestion
		if p.pos < len(p.src) && p.src[p.pos] == '-' {
			p.pos++
		}
		digitStart := p.pos
		for p.pos < len(p.src) && isASCIIDigit(p.src[p.pos]) {
			p.pos++
		}
		// BOLT OPTIMIZATION: break on first non-digit, no TrimSpace needed.
		// Valid selector must have at least one digit after = or =-.
		return p.src[start:p.pos], p.pos > digitStart
```

### Issue 2 of 2
internal/i18n/icuparser/parse_test.go:54-73
**Missing test coverage for edge cases**

The new test exercises `=-1` but leaves two important cases uncovered: (1) a multi-digit negative selector such as `=-10`, confirming the digit loop consumes all digits correctly; and (2) the degenerate `=-` form (minus with no following digit), which should cause `readSelector` to return `false` and produce a parse error. Without the latter, a future refactor of the validity check could silently accept `=-` as a valid selector.

PR #950, opened 2026-06-10
⚡ Bolt: Optimize PHP array parser and marshaler

cungminh2710:
💡 What: Optimized `PHPArrayParser` and `MarshalPHPArrayLocale` in `internal/i18n/translationfileparser/php_array_parser.go`.

🎯 Why: To reduce allocation overhead and improve execution speed for PHP localization files. Significant bottlenecks were identified in string escaping (re-creating replacers), redundant sorting, and byte-by-byte string scanning.

📊 Impact: 
- Marshaling: ~7.7x faster (from ~26ms to ~3.4ms per op for 1k entries) and ~13.8x fewer allocations (from ~23MB to ~1.6MB).
- Parsing: ~12% fewer allocations (from ~14.9k to ~13k allocs).

🔬 Measurement: Verified using benchmarks in `internal/i18n/translationfileparser/php_benchmark_test.go` (created and run during development).
- Before: `BenchmarkPHPArrayParser_Marshal-4 43 26433297 ns/op 23363935 B/op 50961 allocs/op`
- After: `BenchmarkPHPArrayParser_Marshal-4 348 3425686 ns/op 1684319 B/op 17048 allocs/op`

---
*PR created automatically by Jules for task [14141632527323757298](https://jules.google.com/task/14141632527323757298) started by @cungminh2710*


cungminh2710:

---

### Issue 1 of 1
internal/i18n/translationfileparser/php_array_parser.go:72-83
**Implicit ordering invariant now undocumented**

The sort was removed because entries are appended in document (left-to-right) order by the parser, which is correct today. However, this invariant is now silent — if a future refactor causes entries to arrive out of order, `render` will silently skip the displaced entries (the `entry.valueStart < cursor` guard causes them to be dropped) rather than producing an error or panicking. A short comment here would make the assumption explicit and protect against undetected regressions.

PR #953, opened 2026-06-10
🧪 Scout: support optional whitespace in ICU plural offsets

cungminh2710:
💡 What: Added support for optional whitespace in ICU plural offsets and comprehensive unit tests.
🎯 Why: Standard ICU messages often contain spaces for readability, and the parser was previously too restrictive.
🧩 Scope: Modified 'internal/i18n/icuparser/parse.go' and 'internal/i18n/icuparser/parse_test.go'.
✅ Verification: Ran 'go test -v ./internal/i18n/icuparser' and 'go test ./...'.
🛡️ Confidence: High, as it fixes a known parsing limitation and is verified by targeted and regression tests.

---
*PR created automatically by Jules for task [3432802721482294323](https://jules.google.com/task/3432802721482294323) started by @cungminh2710*

### DIFF
--- a/internal/i18n/icuparser/parse.go
+++ b/internal/i18n/icuparser/parse.go
@@ -416,10 +416,48 @@ func (p *astParser) parsePluralOptions() (int, []PluralOption, error) {
 		}
 		p.skipSpaces()
 		// BOLT OPTIMIZATION: Quick check for 'o' or 'O' to avoid strings.EqualFold for most selectors.
-		if len(sel) >= 7 && (sel[0] == 'o' || sel[0] == 'O') && strings.EqualFold(sel[:7], "offset:") {
-			n, err := strconv.Atoi(sel[7:])
+		if (sel[0] == 'o' || sel[0] == 'O') && (strings.EqualFold(sel, "offset") || (len(sel) >= 7 && strings.EqualFold(sel[:7], "offset:"))) {
+			var val string
+			if strings.EqualFold(sel, "offset") {
+				p.skipSpaces()
+				if !p.consume(':') {
+					return 0, nil, fmt.Errorf("expected ':' after offset keyword at %d", p.pos)
+				}
+				p.skipSpaces()
+				numStart := p.pos
+				if p.pos < len(p.src) && p.src[p.pos] == '-' {
+					p.pos++
+				}
+				digitStart := p.pos
+				for p.pos < len(p.src) && isASCIIDigit(p.src[p.pos]) {
+					p.pos++
+				}
+				if p.pos == digitStart {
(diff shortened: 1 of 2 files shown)


cungminh2710:

---

### Issue 1 of 3
internal/i18n/icuparser/parse_test.go:371-408
The expected offset is derived from `tt.name` rather than from the test struct itself. If the test case named `"multiple spaces"` is ever renamed or reordered, the expected value silently flips to `1` without any compile-time or test-definition error. Encode the expectation directly in the table row.

```suggestion
	tests := []struct {
		name           string
		msg            string
		expectedOffset int
	}{
		{
			name:           "no space",
			msg:            "{count, plural, offset:1 one {# item} other {# items}}",
			expectedOffset: 1,
		},
		{
			name:           "space after colon",
			msg:            "{count, plural, offset: 1 one {# item} other {# items}}",
			expectedOffset: 1,
		},
		{
			name:           "space before colon",
			msg:            "{count, plural, offset : 1 one {# item} other {# items}}",
			expectedOffset: 1,
		},
		{
			name:           "multiple spaces",
			msg:            "{count, plural, offset  :  2 one {# item} other {# items}}",
			expectedOffset: 2,
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			elems, err := Parse(tt.msg, nil)
			if err != nil {
				t.Fatalf("Parse() failed: %v", err)
			}
			pl := elems[0].(PluralElement)
			if pl.Offset != tt.expectedOffset {
				t.Errorf("expected offset %d, got %d", tt.expectedOffset, pl.Offset)
			}
		})
	}
```

### Issue 2 of 3
internal/i18n/icuparser/parse.go:439-456
**Redundant `skipSpaces()` in the `else` branch**

`readSelector` already stops at the first whitespace character, and the unconditional `p.skipSpaces()` at line 417 runs immediately after — consuming all trailing whitespace before we even enter this branch. As a result, by the time `val == ""` is tested, `p.pos` already points past any whitespace separating `"offset:"` from the digit string. The inner `p.skipSpaces()` call (line 443) is always a no-op and may mislead future readers into thinking it does meaningful work.

### Issue 3 of 3
internal/i18n/icuparser/parse.go:420-456
**Duplicated digit-reading logic**

The block that reads an optional leading `-` followed by ASCII digits and validates at least one digit was written twice — once for the `strings.EqualFold(sel, "offset")` path (lines 428–438) and again for the `val == ""` path in the `else` branch (lines 445–455). If the digit-parsing logic ever changes (e.g., to support `+`, or to allow multi-segment numbers), one copy is likely to be missed. Extracting this into a small helper (e.g., `readOffsetNumber() (string, error)`) would eliminate the duplication.

PR #956, opened 2026-06-11
⚡ Bolt: optimize ARB marshaling

cungminh2710:
💡 What: Optimized ARB marshaling logic.
🎯 Why: ARB files can be large, and the previous implementation performed redundant sorting and allocations for every key/value pair.
📊 Impact: ~11-18% speed improvement and reduced memory allocations for ARB file generation.
🧪 Measurement: Verified with `GOWORK=off go test -v -bench=BenchmarkARB -benchmem ./internal/i18n/translationfileparser/`.

---
*PR created automatically by Jules for task [12499635053686880269](https://jules.google.com/task/12499635053686880269) started by @cungminh2710*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -152,3 +152,7 @@
 ## 2026-06-10 - Optimizing PHP array parsing and marshaling
 **Learning:** Re-creating `strings.NewReplacer` in a hot loop is extremely expensive due to internal trie construction. Additionally, redundant `slices.Sort` calls on segments that are already in document order and missing `strings.Builder.Grow` hints in renderers are significant avoidable overheads.
 **Action:** Move `strings.NewReplacer` to package-level variables for static rules. Use `strings.Builder.Grow` in renderers. Remove redundant sorting by ensuring parsers produce ordered segments.
+
+## 2026-09-15 - Optimizing ARB marshaling via string fast-paths and partial sorting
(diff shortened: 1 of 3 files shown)


cungminh2710:

---

### Issue 1 of 2
internal/i18n/translationfileparser/arb_parser.go:338
**Silenced error in `marshalJSONString`**

`json.Marshal(s)` on a `string` returns `(nil, error)` only in the unlikely event of invalid UTF-8 (Go replaces bad bytes, so in practice this never errors), but silently discarding the return value with `_` is inconsistent with how all other `json.Marshal` calls in this file are handled. If this ever returned `nil`, the caller would write a raw `nil` slice to the output, producing a corrupted ARB field.

### Issue 2 of 2
internal/i18n/translationfileparser/arb_parser.go:318-328
**`<`, `>`, and `&` are no longer HTML-escaped in ARB values**

`isSimpleJSONString` treats `<` (0x3C), `>` (0x3E), and `&` (0x26) as simple characters because they fall in the 0x20–0x7E range. The previous `json.Marshal` path encoded them as `\u003c`, `\u003e`, and `\u0026` (Go's default HTML-safe mode). Both encodings are valid JSON, and for Flutter ARB files the unescaped form is actually fine, but this is a silent output format change for any translation string that contains those characters. If any downstream tooling or diffing workflow depends on the escaped form, it will start flagging existing files as changed.

PR #961, opened 2026-06-12
🧪 Scout: Fix path resolution edge case for empty localeDir

cungminh2710:
💡 What: Improved path resolution logic and added a regression test for i18n path resolution with empty tokens.
🎯 Why: A pattern starting with {{localeDir}} (e.g., {{localeDir}}/messages.json) would resolve to /messages.json when the source and target locales matched, causing the isSafeRepositoryRelativePath check to fail because of the leading slash.
🧩 Scope: apps/hyperlocalise-web/src/lib/i18n/i18n-pathresolver.ts, apps/hyperlocalise-web/src/lib/i18n/i18n-pathresolver.test.ts
✅ Verification: Ran `pnpm exec vp test src/lib/i18n/i18n-pathresolver.test.ts` and `pnpm exec vp check` in `apps/hyperlocalise-web`.
🛡️ Confidence: High. The fix uses existing normalization utilities and is protected by a new unit test.

---
*PR created automatically by Jules for task [16844340908794113760](https://jules.google.com/task/16844340908794113760) started by @cungminh2710*

### DIFF
--- a/.jules/scout.md
+++ b/.jules/scout.md
@@ -71,3 +71,11 @@
 ## 2025-06-25 - [Path Resolution with Empty Tokens]
 **Learning:** Path resolution patterns using tokens like {{localeDir}} can produce leading slashes if the token is at the start of the pattern and resolves to an empty string (e.g., when source and target locales match). These leading slashes can cause downstream safety checks that expect strictly relative paths to fail.
 **Action:** Always apply repository-relative path normalization after token substitution and before safety validation to ensure consistent handling of relative paths regardless of token resolution.
+
+## 2025-06-25 - [Slash Collapsing in Path Resolution]
+**Learning:**  may only handle leading/trailing slashes and specific segments (like  or ). It might not collapse internal multiple slashes (e.g., ).
+**Action:** When performing token substitution that might result in empty segments, explicitly collapse multiple slashes using a regex (e.g., ) before applying repository-wide normalization.
+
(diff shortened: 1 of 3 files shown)


cungminh2710:

---

### Issue 1 of 2
apps/hyperlocalise-web/src/lib/i18n/i18n-pathresolver.ts:19
The old `while (path.includes("//"))` loop collapsed internal double slashes anywhere in the path. `normalizeRepositoryRelativePath` only strips *leading* slashes — it has no `//` collapsing for segments in the middle. A pattern like `"locales/{{localeDir}}/messages.json"` with matching source/target locales (localeDir = `""`) now resolves to `"locales//messages.json"` instead of `"locales/messages.json"`. The path passes the safety check (empty segment is not `".."` or `"."`), but the returned value contains a double slash that was previously normalised away.

```suggestion
  path = normalizeRepositoryRelativePath(path.replace(/\/+/g, "/"));
```

### Issue 2 of 2
apps/hyperlocalise-web/src/lib/i18n/i18n-pathresolver.test.ts:22-24
Missing regression test for `{{localeDir}}` in a *non-leading* position with matching locales. Calling `resolveSourcePath("locales/{{localeDir}}/messages.json", "en")` — or equivalently `resolveTargetPath("locales/{{localeDir}}/messages.json", "en", "en")` — produces `"locales//messages.json"` with the current implementation. Adding a test case for this pattern would catch the double-slash regression described in the companion comment.

PR #976, opened 2026-06-13
fix(web): materialize TMS projects for Contentful connections

cungminh2710:
## What changed

Contentful connection saves returned `project_not_found` (404) when users selected external TMS projects from the integrations project picker. Those projects are listed from the live TMS provider with encoded IDs like `ext:crowdin:…`, but they were not always persisted in the local `projects` table required by `contentful_connections.project_id`.

This PR adds `ensureOrganizationProjectRecord`, which materializes external TMS projects from the active provider before creating or updating a Contentful connection. Native projects continue to use the existing database lookup path.

## How to test

- Connect an external TMS provider (e.g. Crowdin) in Integrations.
- Open Integrations → Contentful, select a TMS project, fill in space/token/content types, and save.
- Confirm the request succeeds (no 404) and the connection is listed.
- Run:
  - `cd apps/hyperlocalise-web && vp test src/lib/projects/ensure-organization-project.test.ts src/lib/contentful/connections.test.ts`
  - `cd apps/hyperlocalise-web && vp check --fix`

## Checklist

### DIFF
--- a/apps/hyperlocalise-web/drizzle/0047_shallow_mojo.sql
+++ b/apps/hyperlocalise-web/drizzle/0047_shallow_mojo.sql
@@ -1,0 +1,2 @@
+ALTER TABLE "projects" ADD COLUMN "updated_by_user_id" uuid;--> statement-breakpoint
+ALTER TABLE "projects" ADD CONSTRAINT "projects_updated_by_user_id_users_id_fk" FOREIGN KEY ("updated_by_user_id") REFERENCES "public"."users"("id") ON DELETE set null ON UPDATE no action;
--- a/apps/hyperlocalise-web/drizzle/meta/0047_snapshot.json
+++ b/apps/hyperlocalise-web/drizzle/meta/0047_snapshot.json
@@ -1,0 +1,13587 @@
+{
+  "id": "ea9fd18c-75c0-4581-902b-59fb537310cf",
+  "prevId": "ea5ab31b-e7bd-495d-8ba4-167d6bd01809",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.organization_memberships": {
+      "name": "organization_memberships",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "organization_id": {
+          "name": "organization_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "uuid",
+          "primaryKey": false,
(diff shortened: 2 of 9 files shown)


cungminh2710:

---

### Issue 1 of 3
apps/hyperlocalise-web/src/lib/projects/ensure-organization-project.ts:160-172
**Cross-organization project ID collision on upsert**

The `onConflictDoUpdate` targets only `schema.projects.id`, but external IDs (`ext:crowdin:12345`) are derived from the TMS provider's own auto-incremented integers — not scoped to an organization. Two completely separate organizations using different Crowdin workspaces will regularly produce the same encoded ID. When a collision happens, `organizationId` is never updated, so the row stays owned by org A. Meanwhile, `externalProviderCredentialId` in the `set` clause IS overwritten to org B's credential, silently corrupting org A's record. Org B's subsequent lookup (`WHERE organizationId = 'orgB' AND id = 'ext:crowdin:12345'`) finds nothing, re-enters the materialization path, and the credential-corruption cycle repeats. The function still returns `ok(canonicalProjectId)`, so org B's Contentful connection is created pointing to org A's project row, where org A's TMS jobs and jobs-counts queries will now use the wrong credential.

### Issue 2 of 3
apps/hyperlocalise-web/src/lib/projects/ensure-organization-project.ts:188-196
`unwrapOrganizationProjectRecord` throws `new Error(result.error.code)`, which discards all the structured context (`reason`, `organizationId`, `projectId`, `providerKind`, etc.) that `projectNotFound` carefully builds. Any catch boundary upstream will only see the opaque string `"project_not_found"` with no way to distinguish a missing native project from a TMS provider mismatch.

```suggestion
export function unwrapOrganizationProjectRecord(
  result: Result<string, EnsureOrganizationProjectError>,
): string {
  if (isErr(result)) {
    const { code, reason, organizationId, projectId, providerKind, externalProjectId, activeProviderKind } =
      result.error;
    throw Object.assign(new Error(code), {
      reason,
      organizationId,
      projectId,
      providerKind,
      externalProjectId,
      activeProviderKind,
    });
  }

  return result.value;
}
```

### Issue 3 of 3
apps/hyperlocalise-web/src/lib/projects/ensure-organization-project.ts:160-172
The conflict-update path omits `updatedByUserId`, so the audit column is never refreshed when an existing external project row is updated. Every subsequent re-sync of the same project will silently leave the actor as whoever first materialized the record.

```suggestion
    .onConflictDoUpdate({
      target: schema.projects.id,
      set: {
        name: liveProject.name,
        sourceLocale: liveProject.sourceLocale,
        targetLocales: liveProject.targetLocales,
        externalProjectUrl: liveProject.externalProjectUrl,
        isActive: liveProject.isActive,
        externalProviderCredentialId: credential.id,
        updatedByUserId: input.userId ?? null,
        lastSyncedAt: new Date(),
        updatedAt: new Date(),
      },
    });
```

PR #1015, opened 2026-06-15
feat(cli): add crowdin download sources command

cungminh2710:
## What changed

- Added `hyperlocalise crowdin download sources` to download Crowdin source files into configured `files[].source` paths or a single `--output` file via `--file-id`.
- Moved the existing translation download flow under `hyperlocalise crowdin download translations` (breaking change from the flat `crowdin download` command).
- Wired provider agent file translation to download Crowdin sources inside the sandbox with the new CLI command instead of server-side fetches.
- Updated English docs for the new subcommands.

## How to test

- `make fmt && make lint && make test`
- `go run ./apps/cli crowdin download sources --help`
- `go run ./apps/cli crowdin download translations --help`
- `cd apps/hyperlocalise-web && vp test src/lib/translation/sandbox-translation.test.ts src/lib/providers/agent-runs/provider-agent-file-translate.test.ts`
- `cd docs && mint broken-links`

## Checklist

### DIFF
--- a/apps/cli/cmd/crowdin.go
+++ b/apps/cli/cmd/crowdin.go
@@ -31,6 +31,17 @@ type crowdinCommonOptions struct {
 	dryRun                  bool
 }
 
+type crowdinDownloadSourcesOptions struct {
+	configPath   string
+	identityPath string
+	branch       string
+	fileID       int
+	sourcePaths  []string
+	outputPath   string
+	force        bool
+	dryRun       bool
+}
+
 type crowdinGlossaryDownloadOptions struct {
 	configPath   string
 	identityPath string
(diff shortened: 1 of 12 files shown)


cungminh2710:

---

### Issue 1 of 3
apps/hyperlocalise-web/src/lib/providers/agent-runs/provider-agent-file-translate.ts:349-357
**Crowdin context loaded once per source file**

`loadProviderCrowdinDownloadContext` is called inside the `for (const sourceFile of sourceFiles)` loop. The returned `externalProjectId`, `secretMaterial`, and `baseUrl` are project-scoped — identical for every file in the same Crowdin project. This means N database queries and credential decryptions for N source files. The call should be hoisted above the source-file loop and the result reused across iterations.

### Issue 2 of 3
apps/cli/cmd/crowdin.go:282-283
`--output` is silently discarded when `--file-id` is not set. A user who passes `--output /some/path` without `--file-id` will get no error and the flag has no effect, which is surprising. Adding an early guard makes the intent unambiguous.

```suggestion
	outputPath := strings.TrimSpace(o.outputPath)
	if outputPath != "" && o.fileID == 0 {
		return fmt.Errorf("crowdin download sources: --output requires --file-id")
	}
	if o.fileID > 0 {
```

### Issue 3 of 3
apps/hyperlocalise-web/src/lib/translation/sandbox-translation.ts:1332-1352
**`uploadCrowdinTranslationsInSandbox` has no callers in this PR**

The newly exported function is not referenced in any changed file. If it's scaffolding for a follow-up, consider keeping it in a separate commit or PR to avoid accumulating dead exports. If it's needed here, the call site is missing.

PR #1019, opened 2026-06-15
🧪 Scout: add unit tests for XML utility functions

cungminh2710:
💡 What: Added a new unit test file `internal/i18n/translationfileparser/xml_util_test.go` covering XML utility functions.
🎯 Why: The functions in `xml_util.go` (escaping, entity detection, attribute retrieval) were previously untested, posing a risk for XML-based parsers like XLIFF and Android XML.
🧩 Scope: `internal/i18n/translationfileparser/xml_util.go`
✅ Verification: Ran `go test -v ./internal/i18n/translationfileparser/xml_util_test.go ./internal/i18n/translationfileparser/xml_util.go` and `go test ./...`. Verified with `make lint`.
🛡️ Confidence: Ensures that XML escaping and entity handling remain correct as parsers are optimized or expanded.

---
*PR created automatically by Jules for task [3700871914613467272](https://jules.google.com/task/3700871914613467272) started by @cungminh2710*

### DIFF
--- a/internal/i18n/translationfileparser/xml_util_test.go
+++ b/internal/i18n/translationfileparser/xml_util_test.go
@@ -19,9 +19,11 @@ func TestEscapeXMLText(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		if got := escapeXMLText(tt.in); got != tt.want {
-			t.Errorf("escapeXMLText(%q) = %q, want %q", tt.in, got, tt.want)
-		}
+		t.Run(tt.in, func(t *testing.T) {
+			if got := escapeXMLText(tt.in); got != tt.want {
+				t.Errorf("escapeXMLText() = %q, want %q", got, tt.want)
+			}
+		})
 	}
 }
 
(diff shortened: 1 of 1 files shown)


cungminh2710:

---

### Issue 1 of 2
internal/i18n/translationfileparser/xml_util_test.go:55-75
**Missing coverage for loop-continuation path**

`containsXMLTextEntityReference` iterates past unrecognised entities (line 46 of `xml_util.go`: `searchFrom = entityStart`) so it can still return `true` when a valid entity appears later in the string. No test exercises this path. A case like `{"&unknown; &amp;", true}` would validate that the loop continues correctly; without it, the only tested non-trivial entity cases are single-entity strings, leaving the multi-entity continuation logic silently untested.

### Issue 2 of 2
internal/i18n/translationfileparser/xml_util_test.go:8-104
**Sub-tests not wrapped with `t.Run`**

All four test functions use a raw `for` loop with `t.Errorf`. This means a failing sub-case is identified only by the printed argument values, not by a named sub-test path. Using `t.Run(tt.name, ...)` (or an inline label derived from `tt.in`) would let `go test -run TestEscapeXMLText/hello` target a specific case and would make CI failure output much easier to read. The test table in `TestAttrValue` already has a `name` field that is populated but never passed to `t.Run`, which suggests this was the intended pattern.

PR #1027, opened 2026-06-16
⚡ Bolt: optimize Java Properties and Apple Strings parsers

cungminh2710:
### 💡 What:
Optimized the Java Properties and Apple Strings parsers by replacing O(N²) line counting with O(N) incremental tracking and implementing an allocation-free fast-path for simple property lines.

### 🎯 Why:
The previous implementation called `lineNumberAt` for every entry/token, which scans the entire file prefix each time, leading to quadratic time complexity on large files. Additionally, `JavaPropertiesParser` was always allocating buffers and mapping slices even for simple single-line properties.

### 📊 Impact:
- **Java Properties Parser:** ~2.5x faster, ~80% fewer allocations.
- **Java Properties Marshaler:** ~2x faster, ~65% fewer allocations.
- **Apple Strings Parser:** Fixed O(N²) bottleneck for large files.

### 🔬 Measurement:
Verified with `internal/i18n/translationfileparser/properties_benchmark_test.go` and existing unit tests in `properties_parser_test.go` and `strings_parser_test.go`.

---
*PR created automatically by Jules for task [17995218078642072213](https://jules.google.com/task/17995218078642072213) started by @cungminh2710*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -63,6 +63,14 @@
 ## 2026-06-20 - Optimizing XCStrings path and key construction
 **Learning:** XCStrings parsing involves frequent recursive construction of path labels (e.g., `strings.KEY.localizations.LOCALE`). Using `fmt.Sprintf` for these labels in loops or recursion introduces significant reflection overhead. String concatenation is a much more efficient alternative.
 **Action:** Replaced `fmt.Sprintf` with string concatenation in `internal/i18n/translationfileparser/xcstrings_parser.go`, resulting in ~10% faster parsing and ~14% fewer allocations.
+
+## 2026-07-28 - Eliminating O(N²) line-number counting in sequential parsers
+**Learning:** Repeatedly calling a function that scans the entire prefix of a file to count newlines (e.g., `lineNumberAt(text, offset)`) inside a parser loop leads to $O(N^2)$ complexity. For sequential parsers, tracking a `currentLine` counter incrementally as the parser advances is significantly more efficient ($O(N)$).
+**Action:** Updated `JavaPropertiesParser` and `AppleStringsParser` to track line numbers incrementally, avoiding massive slowdowns on large localization files.
+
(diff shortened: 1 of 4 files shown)


cungminh2710:

---

### Issue 1 of 3
internal/i18n/translationfileparser/properties_parser.go:231
Negative capacity panic in `render`. `seen` is populated from every template entry, while `values` only contains the caller-supplied keys. When the template has more keys than `values` (the common case, e.g. a fully-translated template being re-marshalled with just a few updates), `len(seen) > len(values)`, making the expression negative — and `make([]string, 0, negative)` panics at runtime with "makeslice: cap out of range".

```suggestion
	missingCap := len(values) - len(seen)
	if missingCap < 0 {
		missingCap = 0
	}
	missing := make([]string, 0, missingCap)
```

### Issue 2 of 3
internal/i18n/translationfileparser/properties_parser.go:300-301
Inaccurate error line for multi-line continuations. `lineNumber` is the line of the first physical line of the logical line, but this error fires at whatever later physical line has the dangling continuation. The old code used `lineNumberAt(text, contentEnd)` which gave the correct current line. For a 3-physical-line logical line, the error now says "line 1" when the actual offending backslash is on line 3.

### Issue 3 of 3
internal/i18n/translationfileparser/properties_parser.go:258-280
Redundant physical-line scan in the slow path. `readPropertiesPhysicalLine(text, start)` is called before the fast-path check and then called again for the same `start` in the first loop iteration (`pos = start`). This doubles the work for every continued logical line. The result from the pre-check could be threaded into the loop to avoid the second scan.

cungminh2710:
@jules The fast-path rawStart+valueStart mapping and the boundaryRaw index arithmetic are consistent with how the rest of the parser uses these offsets. The one remaining issue is that missingCap can collapse to zero when the template has more entries than the provided values map, causing the missing slice to reallocate on every growth step rather than being pre-sized; this is a performance concern rather than a correctness defect.

The missingCap calculation in properties_parser.go's render method warrants a second look.

PR #1030, opened 2026-06-16
🧪 Scout: fix XLIFF entity and CDATA decoding

cungminh2710:
- 💡 What: Added `TestXLIFFParserDecodesEntities` to `xliff_parser_test.go` and updated `normalizeXLIFFInternalMarkup` in `xliff_parser.go`.
- 🎯 Why: The previous XLIFF parser implementation used a slicing approach that preserved internal tags but failed to decode XML entities (e.g., `&amp;`) or CDATA sections. This led to issues where raw entities were passed to translators or caused ICU parsing failures.
- 🧩 Scope: `internal/i18n/translationfileparser/xliff_parser.go` and `internal/i18n/translationfileparser/xliff_parser_test.go`.
- ✅ Verification: Ran `go test -v ./internal/i18n/translationfileparser` which confirmed the new test passes and no regressions were introduced. Verified correct tag ordering in mixed content.
- 🛡️ Confidence: High. The fix uses a robust `xml.Decoder` loop with explicit flushing to maintain ordering, and includes a safe fallback to the original string on decoding errors.

---
*PR created automatically by Jules for task [2884346538768286736](https://jules.google.com/task/2884346538768286736) started by @cungminh2710*

### DIFF
--- a/.jules/scout.md
+++ b/.jules/scout.md
@@ -73,16 +73,8 @@
 **Action:** Always apply repository-relative path normalization after token substitution and before safety validation to ensure consistent handling of relative paths regardless of token resolution.
 
 ## 2025-06-25 - [Slash Collapsing in Path Resolution]
-**Learning:**  may only handle leading/trailing slashes and specific segments (like  or ). It might not collapse internal multiple slashes (e.g., ).
-**Action:** When performing token substitution that might result in empty segments, explicitly collapse multiple slashes using a regex (e.g., ) before applying repository-wide normalization.
-
-## 2025-06-25 - [Slash Collapsing in Path Resolution]
 **Learning:** `normalizeRepositoryRelativePath` may only handle leading/trailing slashes and specific segments (like `.` or `..`). It might not collapse internal multiple slashes (e.g., `a//b`).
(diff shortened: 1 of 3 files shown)


cungminh2710:

---

### Issue 1 of 2
.jules/scout.md:80-83
**Duplicate learning entry with stripped formatting**

The same "Mixed XML Encoder and Buffer Writes" learning is added twice — the first copy (lines 80-83) has all its inline code references stripped (backtick-quoted terms appear as empty spans), leaving unreadable prose. The second copy (lines 85-88) is the correct version. The first entry should be removed to keep the scout log clean and readable.

### Issue 2 of 2
internal/i18n/translationfileparser/xliff_parser.go:163-167
**Unescaped CharData produces non-XML-valid mixed-content output**

When a `<source>` element contains both entities and inline markup (e.g., `Value &amp; <ph id="1"/> markup`), the new code writes decoded `CharData` bytes directly to `out` while inline element tokens go through the XML encoder. The resulting string — `Value & <ph id="1"></ph> markup` — is not well-formed XML: the bare `&` will cause any downstream XML parser to reject the segment. The previous implementation preserved entities in the encoder path, keeping the output valid XML throughout. Any XLIFF-processing step that re-parses these values (e.g., a segment splitter, an inline-markup extractor, or a future marshal path) will need to handle this new unescaped format explicitly.

PR #1045, opened 2026-06-17
🧪 Scout: improve custom tag identification for structural parity

cungminh2710:
This change improves the reliability of HTML tag structural parity checks by ensuring custom tags common in modern web frameworks (like MDX and Web Components) are correctly identified. Previously, tags containing dots, underscores, or digits were often ignored, which could lead to missed mismatches if they were removed in translations.

Changes:
- Updated `internal/i18n/htmltagparity/htmltagparity.go` to include `.`, `_`, and digits in the tag identification logic.
- Added new test cases to `internal/i18n/htmltagparity/htmltagparity_test.go` covering these new tag patterns.
- Updated Scout's journal with findings.

---
*PR created automatically by Jules for task [12203064222350246634](https://jules.google.com/task/12203064222350246634) started by @cungminh2710*


cungminh2710:

---

### Issue 1 of 2
internal/i18n/htmltagparity/htmltagparity.go:115
**Potential false positives for path/template placeholder tokens**

`strings.ContainsAny(tag, "-:._0123456789")` now treats any tag whose name contains a digit, dot, or underscore as structural markup. In technical documentation it is common to have template-style placeholders like `<resource_id>`, `<endpoint_v1>`, or `<base_url>` that are written with angle brackets but are *not* structural HTML/MDX. Prior to this change those tokens fell through to `rawTagHasAttributes`, which returns `false` for attribute-less tokens and thus did not flag them. With this change, a translation that rephrases such a sentence without the placeholder token (e.g. a translator converts `` see `api/<version_1>` `` to flowing prose that omits the placeholder) will now be incorrectly flagged as a structural mismatch.

### Issue 2 of 2
internal/i18n/htmltagparity/htmltagparity_test.go:43-49
**`TestMismatchIgnoresAngleBracketPathTokens` does not cover asymmetric removal**

The test comment says "expected `<name>` in paths not to count as HTML tags", but the assertion passes whether `<name>` is treated as markup or not — because `<name>` appears identically in both `src` and `tgt`, so the normalised tag lists are always equal. The test would also need a case where the translation *omits* the path token (e.g. `tgt = "Sử dụng 'bedrock' trong 'llm.profiles.provider'."`) to actually verify that path-like tokens are ignored. The same gap now applies to the newly included characters: there is no test covering a path/placeholder token containing `_`, `.`, or a digit that a translator legitimately drops.

PR #1057, opened 2026-06-19
feat(web): improve CAT translation assistance UX

cungminh2710:
## What changed

Improves the CAT workspace translation assistance experience:

- **TM matches**: one-click **Use** in the intelligence panel, single color-coded match badge (green for 100%/context, yellow for fuzzy), confirmation below 70%, and optional auto-fill from 100% TM on empty targets
- **Concordance**: loads on segment focus (not only during AI review) and uses the project's linked Crowdin credential instead of the org's most recently updated credential
- **Unsaved changes**: dirty indicators in the editor and queue, navigation/page-change confirmation, and browser `beforeunload` guard
- **Editor toolbar**: copy source and clear target actions
- **maxLength**: live character counter in the target editor
- **Cleanup**: removed unused `mt` suggestion source type

## How to test

- Open a CAT file workspace with TM/glossary concordance enabled
- Select a segment and confirm glossary/TM load without clicking **Get recommendation**
- Apply a TM match via **Use**; verify fuzzy matches below 70% prompt for confirmation
- Edit a target, navigate to another segment or queue page, and confirm the unsaved-changes dialog
- Use **Copy source** / **Clear target** in the target toolbar
- For segments with `maxLength`, confirm the live `{count}/{maxLength}` counter updates while typing
- Run `cd apps/hyperlocalise-web && vp test src/components/cat/ && vp check --fix`

## Checklist

### DIFF
--- a/apps/hyperlocalise-web/src/api/routes/project/project-cat.route.test.ts
+++ b/apps/hyperlocalise-web/src/api/routes/project/project-cat.route.test.ts
@@ -204,7 +204,6 @@ describe("project file CAT routes", () => {
       expect.objectContaining({
         projectId: "ext:crowdin:42",
         providerKind: "crowdin",
-        externalProjectId: "42",
         sourceLocale: "en",
         targetLocale: "fr",
         sourceText: "Hello workspace",
--- a/apps/hyperlocalise-web/src/api/routes/project/project.route.ts
+++ b/apps/hyperlocalise-web/src/api/routes/project/project.route.ts
@@ -697,7 +697,6 @@ export function createProjectRoutes(options: CreateProjectRoutesOptions = {}) {
             organizationId: c.var.auth.organization.localOrganizationId,
             projectId: params.projectId,
             providerKind: target.kind === "provider" ? target.providerKind : null,
-            externalProjectId: target.kind === "provider" ? target.externalProjectId : null,
             sourceLocale: body.sourceLocale,
             targetLocale: body.targetLocale,
             sourceText: body.sourceText,
--- a/apps/hyperlocalise-web/src/components/cat/cat-dirty-state.test.ts
+++ b/apps/hyperlocalise-web/src/components/cat/cat-dirty-state.test.ts
@@ -1,0 +1,83 @@
+import { describe, expect, it } from "vite-plus/test";
+
+import {
+  buildSavedTargetTextMap,
+  collectDirtySegmentIds,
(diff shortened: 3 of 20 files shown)


cungminh2710:

---

### Issue 1 of 3
apps/hyperlocalise-web/src/components/cat/cat-intelligence-panel.tsx:329-337
**`matchKindExact` / `matchKindFuzzy` messages are dead code**

`tmMatchBadgeLabel` only handles `"context"` and falls through to the `matchPercent` format string for everything else, so `matchKindExact` and `matchKindFuzzy` are never rendered. An "exact" (100%) match badge therefore shows `"100%"` instead of `"100% match"`, which diverges from the PR description. Either use the messages in the function or remove the unused definitions.

### Issue 2 of 3
apps/hyperlocalise-web/src/components/cat/cat-workspace-container.tsx:858-880
**TM auto-fill creates instant dirty state on every first-focus**

When concordance loads and a 100 % TM match is found for an empty target, `updateSegmentTarget` / `onTargetChange` are called but `markSegmentTargetSaved` is not. The segment immediately registers as dirty, showing the "Unsaved" badge and blocking navigation with a confirmation dialog — even though the user has not typed anything. Users who open segments in sequence (e.g. reviewing an import batch) will be confronted with the leave-without-saving dialog on every segment they visit that was auto-filled. Consider calling `markSegmentTargetSaved(saved, segmentId, bestTmMatch.targetText)` after the auto-fill, or surface the auto-fill differently (e.g. as a dismissible suggestion) so the dirty state only appears after an explicit user edit.

### Issue 3 of 3
apps/hyperlocalise-web/src/lib/translation/load-cat-segment-concordance.ts:145-169
**`externalProjectId` in the public interface is now a no-op value**

The Crowdin branch is still gated on `input.externalProjectId` being truthy (line 154), but `loadCrowdinLiveConcordance` no longer uses the passed-in value — it re-fetches `externalProjectId` from the database via `loadCrowdinProjectCredential`. Any caller that passes a different `externalProjectId` from what is stored on the project row will silently have that value ignored. More importantly, if the DB query returns `null` (e.g. `externalProviderCredentialId` not set on the project row), the Crowdin branch returns empty arrays and does **not** fall through to the generic TM path — even though `providerKind === "crowdin" && externalProjectId` was truthy. Consider removing `externalProjectId` from the input type (it is no longer read), or at minimum add a code comment clarifying that the parameter is used only as a feature gate.

PR #1066, opened 2026-06-19
⚡ Bolt: precompute HTML syntax counts and optimize scanning

cungminh2710:
### 💡 What
This optimization improves the performance of the i18n rendering and parsing logic, specifically for HTML, Markdown, and Liquid files. 

Key changes include:
- Precomputing the `sourceSyntaxCount` for each translatable segment during the parsing phase and storing it in the `htmlPart` and `markdownPart` structs.
- Updating `IntroducesRawHTMLSyntax` to take the precomputed count, avoiding a redundant $O(N)$ scan of the source string during every render call.
- Optimizing `rawHTMLSyntaxStartCount` to use `strings.IndexByte` for faster character scanning.
- Implementing an allocation-free `isAllHTMLWhitespace` check that operates directly on byte/string data without intermediate `strings.TrimSpace` allocations.
- Adding capacity hints to `strings.Builder` (using `Grow`) and internal slices/maps in the HTML parser to minimize re-allocations.
- Deferring string conversions in the `html.Tokenizer` loop.

### 🎯 Why
Document rendering is a high-frequency operation. In the original implementation, the `IntroducesRawHTMLSyntax` function scanned the source string for every translatable segment to count HTML-like characters, leading to avoidable CPU cycles. Furthermore, the parser was performing several intermediate string allocations for whitespace checks and literal content that could be handled more efficiently with byte-level operations and capacity hinting.

### 📊 Impact
- **Faster Rendering:** The render path for HTML, Markdown, and Liquid is now faster by avoiding redundant $O(N)$ source scans.
- **Lower Memory Pressure:** Reduced heap allocations during parsing by using byte-level checks and pre-allocating slice/map capacities.
- **Efficient Scanning:** `rawHTMLSyntaxStartCount` is significantly faster due to the use of Go's optimized `strings.IndexByte`.

### 🔬 Measurement
Verified the improvements using local benchmarks (`BenchmarkHTMLParser` and `BenchmarkHTMLMarshal`) and ensured functional correctness by running the full unit test suite (`go test ./...`). Structural refactoring was verified across the CLI and core library packages.

---
*PR created automatically by Jules for task [12835489267583032486](https://jules.google.com/task/12835489267583032486) started by @cungminh2710*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -166,6 +166,10 @@
 **Learning:** Re-creating `strings.NewReplacer` in a hot loop is extremely expensive due to internal trie construction. Additionally, redundant `slices.Sort` calls on segments that are already in document order and missing `strings.Builder.Grow` hints in renderers are significant avoidable overheads.
 **Action:** Move `strings.NewReplacer` to package-level variables for static rules. Use `strings.Builder.Grow` in renderers. Remove redundant sorting by ensuring parsers produce ordered segments.
 
+## 2026-10-05 - Precomputing syntax counts and optimizing scanning
+**Learning:** Repeatedly scanning the same source strings for syntax validation (e.g., `IntroducesRawHTMLSyntax`) during rendering creates $O(N)$ overhead per segment that can be avoided by precomputing counts during parsing. Additionally, manual byte-by-byte loops for character searching (like finding '<') are significantly slower than Go's optimized `strings.IndexByte`.
+**Action:** Precompute `sourceSyntaxCount` during parsing for HTML/Markdown parts and store it in the struct. Optimize `rawHTMLSyntaxStartCount` using `strings.IndexByte` for faster character discovery.
+
 ## 2026-09-15 - Optimizing ARB marshaling via string fast-paths and partial sorting
 **Learning:** For JSON-based formats like ARB, bypassing `json.Marshal` for simple ASCII strings and avoiding full map sorts when only a few keys are new provides significant efficiency gains. Heuristic capacity hints for maps and slices also minimize GC pressure during large file processing.
 **Action:** Implemented `isSimpleJSONString` fast-path and refactored `MarshalARB` to sort only new keys, resulting in ~11-18% speedup and reduced allocations.
--- a/apps/cli/internal/i18n/runsvc/translation_output_validate.go
+++ b/apps/cli/internal/i18n/runsvc/translation_output_validate.go
@@ -52,7 +52,7 @@ func validateTranslatedOutputForKind(kind translationOutputKind, source, transla
 		if err := translationfileparser.ValidateMarkdownInternalPlaceholders(source, translated); err != nil {
 			return &postTranslateValidationError{msg: err.Error()}
 		}
-		if translationfileparser.IntroducesRawHTMLSyntax(source, translated) {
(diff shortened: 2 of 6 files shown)


cungminh2710:

---

### Issue 1 of 2
internal/i18n/translationfileparser/html_parser.go:291-325
**`z.Raw()` called after `z.TagName()` in the non-skip paths**

The Go `html.Tokenizer` documentation explicitly warns: *"Calling Next, Token, TagName or TagAttr may change the contents of the returned slice"* from `Raw()`. The original code addressed this with an explicit comment and captured `raw := string(z.Raw())` **before** any `TagName()` call. The PR correctly preserves that pattern in the `skipDepth > 0` block (lines 269–283: `rawBytes := z.Raw()` precedes `z.TagName()`), but drops it in the non-skip switch cases where `z.TagName()` is called first and `z.Raw()` follows. If the tokenizer's internal buffer is ever reallocated or the `raw` window is adjusted during `TagName()` parsing, the bytes written to the output document will be wrong (corrupted or empty tag literals). The inconsistency between the two paths within the same function is a strong signal that the non-skip paths are not handling this correctly.

### Issue 2 of 2
internal/i18n/translationfileparser/html_parser.go:101-106
`isAllHTMLWhitespace` omits form feed (`\f`, 0x0C) from its ASCII fast-path. HTML5 defines ASCII whitespace as space, tab, LF, **FF (form feed)**, and CR — the original `strings.TrimSpace` treated all of these as whitespace. With the new function, a text token containing only `\f` characters hits the ASCII branch, returns `false`, and is forwarded to `handleTranslatable` instead of being appended as a literal. This diverges from both the original behaviour and the HTML5 whitespace definition.

```suggestion
func isAllHTMLWhitespace(s string) bool {
	for i := 0; i < len(s); {
		if s[i] < 0x80 {
			if s[i] != ' ' && s[i] != '\t' && s[i] != '\n' && s[i] != '\r' && s[i] != '\f' {
				return false
			}
```

cungminh2710:

---

### Issue 1 of 1
internal/i18n/translationfileparser/markdown_md_parser.go:77-85
**Missing `sourceSyntaxCount` in AST markdown parser**

`parseMarkdownASTDocument` (used for all `.md` files) creates `markdownPart` structs without setting `sourceSyntaxCount`, so the field defaults to `0`. Because `renderMarkdownPartWithDiagnostics` now calls `IntroducesRawHTMLSyntax(part.sourceSyntaxCount, rendered)`, the baseline for AST-parsed parts is always `0` instead of the actual count from `part.source`.

If the source `placeholdered` string contains a `<` sequence that `protectStandardMarkdownInlineSyntax` does not protect — specifically `<>`, `<!`, or `<?` — then `rawHTMLSyntaxStartCount(placeholdered) > 0`. A translation that legitimately reflects those same characters back would now fail the check and trigger an incorrect source fallback, because `0` is used as the baseline instead of the true source count.

The three `emitMarkdownLineParts` / `emitFrontmatterLineParts` call sites in `markdown_parser.go` were updated to carry `sourceSyntaxCount`, but the equivalent place in `markdown_md_parser.go` was not.

cungminh2710:
@cursoragent The one open concern — rawBytes from z.Raw() may point into the tokenizer's mutable internal buffer and is used after z.TagName() has been called in the non-skip switch cases

PR #1085, opened 2026-06-21
feat(canva-app): upload design file and sync localized text

cungminh2710:
## Summary

Implements the Hyperlocalise Canva localization workflow and moves the backend into `hyperlocalise-web` with per-org Canva connections.

### Canva app (`apps/canva-app`)

- Extract text from selected pages, upload as JSON, run translation, sync back
- Multi-page support via `openDesign({ type: "all_pages" })`
- Connection-token based auth (no separate Express backend)
- Client polls `GET /api/integrations/canva/localize/:jobId` after starting a job

### Web API (`apps/hyperlocalise-web`)

- **`canva_connections` table** — per-org connection bound to an API key, project, and default locales
- **Org routes** — `POST/GET/PATCH/DELETE /api/orgs/:slug/canva-connections` (+ regenerate token)
- **Integration routes** — `/api/integrations/canva/localize` with:
  - `POST /localize` — uploads file and enqueues job (returns `202` with `jobId`)
  - `GET /localize/:jobId` — polls job status and returns translations when complete
  - Explicit CORS allowlist (Canva origins + `CANVA_CORS_ORIGINS`)
  - Connection token auth (`X-Hyperlocalise-Connection-Token`)
  - Optional Canva JWT verification (`CANVA_APP_ID`) with cached JWKS client
  - Atomic brand binding with unique-constraint handling
  - Usage tracking even when JWT is absent

### Code review fixes

- Async job flow avoids blocking HTTP requests for up to 180s
- JWKS client cached per `appId` for cross-request signing-key reuse
- Brand binding race closed with transaction + `23505` handling
- Canva app workflow shows uploading step before translation polling
- Translation output JSON parse failures return stable error code

### Setup flow

1. In Hyperlocalise, create a Canva connection (pick API key + project + locales)
2. Copy the one-time connection token into the Canva app settings
3. Set `CANVA_BACKEND_HOST` to your Hyperlocalise web origin

### Tests

- `vp test` — 1934 passed (web), 11 passed (canva-app)
- `vp check --fix` — clean

<div><a href="https://cursor.com/agents/bc-30e406c8-e7d7-4bce-be5a-745a6328b935"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-30e406c8-e7d7-4bce-be5a-745a6328b935"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


cungminh2710:
@cursoragent Fix the following 6 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 6
apps/hyperlocalise-web/src/lib/canva/localize-design.ts:101-117
**Synchronous polling blocks the HTTP request for up to 180 seconds**

`waitForTranslationJob` holds an open HTTP connection while polling the database every 1.5 s for up to 120 iterations (= 180 s max). When this is called from the `/api/integrations/canva/localize` route handler, the Canva iframe waits for the entire translation job synchronously. Next.js serverless function execution limits (typically 10–60 s depending on the plan and region) will cut this off before the job finishes, returning a 502 to the Canva app mid-job. Even without a timeout, holding a DB connection and an open HTTP connection for minutes is fragile: a dropped keep-alive or Canva's own XHR timeout will silently discard a completed result.

The standard remedy for long-running work is to return a job reference immediately and let the client poll (or use a webhook/SSE) for the result.

### Issue 2 of 6
apps/hyperlocalise-web/src/lib/canva/auth.ts:7-13
**JWKS client is recreated on every request, making the in-memory cache ineffective**

`createJwksClient` is called inside `verifyCanvaToken`, so every authenticated request gets a brand-new `jwksClient` instance with an empty cache. The `cache: true` option only caches signing keys within a single instance, so it provides zero cross-request benefit here. Under any non-trivial load, every request that carries a Canva JWT will fetch Canva's JWKS endpoint. Canva's JWKS endpoint will rate-limit the caller, causing `getSigningKey` to throw and all JWT verifications to fail with a 401 for legitimate users.

The client should be constructed once at module scope (or cached by `appId` in a `Map`) so the signing-key cache is actually reused across requests.

### Issue 3 of 6
apps/hyperlocalise-web/src/lib/canva/connections.ts:273-306
**Race condition in brand binding can produce an unhandled DB unique-constraint error**

`bindCanvaConnectionBrand` does a read-then-conditional-write without a transaction. Two concurrent `POST /localize` requests from the same brand (e.g., two browser tabs) can both pass the `existingBrandConnection` check (neither has bound the brand yet), then both attempt the `UPDATE … WHERE canvaBrandId IS NULL`. When they target *different* connections, the second UPDATE succeeds and sets the same `canvaBrandId` on a second row, violating `canva_connections_org_brand_key`. The DB throws a unique-constraint error that propagates through `localizeCanvaDesign` and the integration route's outer `catch`, which re-throws anything that isn't `"canva_brand_already_bound"`, resulting in a 500 to the user.

Wrapping the check and update in a single transaction (or using `UPDATE … RETURNING` with an `ON CONFLICT` guard) would close this window.

### Issue 4 of 6
apps/canva-app/src/intents/design_editor/app.tsx:209-210
**`"uploading"` step is never rendered — two synchronous state updates are batched by React**

`setWorkflowStep("uploading")` is immediately overwritten by `setWorkflowStep("translating")` within the same synchronous execution frame. React 18 batches both updates into a single render, so the progress indicator jumps straight from `"extracting"` to `"translating"` and the "Upload file" badge is never marked active.

```suggestion
      setWorkflowStep("uploading");
      // Upload step is part of the `localizeDesign` call below; advance to translating after it returns.
      setWorkflowStep("translating");
```

### Issue 5 of 6
apps/hyperlocalise-web/src/api/routes/canva-integration/canva-integration.route.ts:68-88
**`touchCanvaConnectionUsage` is never called when `canvaUser` is absent**

The usage-tracking call only happens inside the `if (canvaUser)` block. When `CANVA_APP_ID` is not set or the Canva user token is missing (both are optional), `canvaUser` is `undefined` and `lastUsedAt` is never updated for the connection. Any connection configured without JWT verification will always show `lastUsedAt: null` regardless of how often it is used.

### Issue 6 of 6
apps/hyperlocalise-web/src/lib/canva/localize-design.ts:311
**`JSON.parse` on translation file content is not guarded against malformed data**

If `getStoredFileContent` returns a buffer that is not valid JSON (empty file, partial write, encoding mismatch), `JSON.parse` throws a `SyntaxError`. This propagates through `localizeCanvaDesign` and is caught by the route handler, which returns a 502 with the raw `SyntaxError` message (e.g., `"Unexpected token …"`), which is not actionable for the Canva app.

```suggestion
    let parsed: Record<string, unknown>;
    try {
      parsed = JSON.parse(content.toString("utf8")) as Record<string, unknown>;
    } catch {
      throw new Error("translation_output_parse_failed");
    }
```

PR #1093, opened 2026-06-21
⚡ Bolt: optimize Liquid parser performance

cungminh2710:
💡 What: Optimized the Liquid parser by migrating from string-based processing to byte-level scanning and reducing allocations in hot paths (token generation, hashing, and tag name extraction).

🎯 Why: The Liquid parser was performing an initial copy of the entire file into a string and then iterating byte-by-byte, creating significant GC pressure and CPU overhead for large templates.

📊 Impact:
- Speed: ~18% faster parsing (from 3.17ms to 2.60ms per op in corpus benchmarks).
- Memory: ~11% reduction in bytes allocated per operation.

🔬 Measurement: Verified with `go test -bench BenchmarkLiquidParserCanonicalCorpus ./internal/i18n/translationfileparser/` and ensured zero regressions with the full Liquid test suite.

---
*PR created automatically by Jules for task [12209560386323857346](https://jules.google.com/task/12209560386323857346) started by @cungminh2710*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -189,3 +189,7 @@
 ## 2026-10-10 - Optimizing Apple Stringsdict parser and renderer
 **Learning:** XML-based pluralization formats like stringsdict benefit significantly from document-order processing. Since the XML decoder visits tokens sequentially, entries can be collected in order, allowing the renderer to bypass expensive sorting and cloning. Additionally, heuristic capacity hints for stacks and maps in recursive-like XML structures (nested dicts) reduce GC pressure.
 **Action:** Removed redundant sorting/cloning in `render` and implemented capacity hints in `parseStringsdictDocument` and helpers in `internal/i18n/translationfileparser/stringsdict_parser.go`.
+
+## 2026-10-15 - Optimizing Liquid parser via byte-level scanning and allocation reduction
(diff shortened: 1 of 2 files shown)


cungminh2710:

---

### Issue 1 of 3
internal/i18n/translationfileparser/liquid_parser.go:309-310
**Stale optimization comment**

The comment says "Use `bytes.Index` to skip ahead" but the function was rewritten to call `bytes.HasPrefix`, not `bytes.Index`. No `bytes.Index` call exists anywhere in `findLiquidDelimiterEnd`. The comment should describe the actual change (converting the string comparison `input[i:i+2] == close` to the equivalent `bytes.HasPrefix(input[i:], close)` to operate directly on `[]byte` without string conversion).

### Issue 2 of 3
internal/i18n/translationfileparser/liquid_parser.go:218-225
**Fast-path skips bulk-write when no special characters remain**

When `bytes.IndexAny(content[i:], "{<")` returns `-1` (no `{` or `<` exists in the rest of the file), the guard `next > 0` prevents writing the remaining bytes in bulk. The loop then falls back to writing each remaining byte individually via `out.WriteByte(ch)` in subsequent iterations. This is the exact "long stretch of plain text with no delimiters" case that would benefit most from the optimisation — a common situation near the end of a template. Handling `next == -1` explicitly with a bulk `out.Write` + break would complete the optimisation.

### Issue 3 of 3
internal/i18n/translationfileparser/liquid_parser.go:248-253
**Stack buffer only avoids heap allocation for short literals**

`buf [64]byte` is declared on the stack and passed as the initial backing to `strconv.AppendInt`. Once the accumulated slice (`index digits + 1 colon + literal`) exceeds 64 bytes, `append(hInput, literal...)` must reallocate on the heap. A Liquid variable like `{{ product.metafields.namespace.some_long_key | default: "fallback" }}` (~75 chars) already crosses that threshold. The code is correct in all cases, but the comment "stack buffer for hashing" overstates the guarantee — it should note that it only avoids the heap allocation for literals shorter than roughly `64 - len(strconv.Itoa(index)) - 1` bytes.

PR #1111, opened 2026-06-23
🧪 Scout: add translator request validation tests

cungminh2710:
- 💡 What: Added dedicated unit tests for `validateRequest`, `validateImageEditRequest`, and `normalizeProvider` in the translator package.
- 🎯 Why: These core validation functions enforce the API contract for all translation and image editing providers, but lacked explicit coverage.
- 🧩 Scope: `internal/i18n/translator/types_test.go` and `.jules/scout.md`.
- ✅ Verification: Ran `make lint` and `go test ./internal/i18n/translator/...` (all passed).
- 🛡️ Confidence: Prevents regressions in mandatory field enforcement and format validation for the translation system.

---
*PR created automatically by Jules for task [18151390689152167708](https://jules.google.com/task/18151390689152167708) started by @cungminh2710*

### DIFF
--- a/.jules/scout.md
+++ b/.jules/scout.md
@@ -95,3 +95,7 @@
 ## 2025-07-20 - [ICU Element Type and Nesting Validation]
 **Learning:** ICU `PluralElement` can represent both `plural` and `selectordinal` types. Structural parity checks rely on the `Type()` method, which correctly chooses the type based on the `Ordinal` flag or an explicit override. Additionally, pound signs (`#`) must be identified as `PoundElement` even when nested inside non-plural blocks (like `select`) if they are ultimately contained by a `plural` or `selectordinal` block.
 **Action:** Always test the `Type()` method for all AST elements, especially for polymorphic elements like `PluralElement`. Ensure nesting tests cover cases where markers like `#` are separated from their parent block by other types of ICU blocks.
+
+## 2026-06-23 - [Translator Request Validation Safety]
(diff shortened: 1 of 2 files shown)


cungminh2710:

---

### Issue 1 of 3
internal/i18n/translator/types_test.go:33-60
**Whitespace-only inputs not tested for `TargetLanguage` and `Model`**

`validateRequest` uses `strings.TrimSpace` on all three fields, yet only the `Source` field has a whitespace-only test case. The tests for `TargetLanguage` and `Model` only cover the empty-string case. If someone later removes the `TrimSpace` call for those fields, no test will catch the regression. The same gap exists in `TestValidateImageEditRequest` for `TargetLanguage`, `Model`, and `Prompt`.

### Issue 2 of 3
internal/i18n/translator/types_test.go:105-115
**`SourceImage: []byte{}` (empty slice) not tested**

`validateImageEditRequest` rejects a source image when `len(req.SourceImage) == 0`, which covers both `nil` and an empty slice. Only the `nil` case is exercised here. An explicit `[]byte{}` test would ensure both forms of "no data" are consistently rejected if the check is ever changed (e.g., to `req.SourceImage == nil`).

### Issue 3 of 3
internal/i18n/translator/types_test.go:63-67
The `tc := tt` copy is a pre-Go-1.22 workaround for the loop-variable capture bug. Since Go 1.22, each iteration gets its own variable, making this copy redundant. All three test loops in this file have the same pattern.

```suggestion
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			t.Parallel()
			err := validateRequest(tt.req)
```

PR #1121, opened 2026-06-24
⚡ Bolt: optimize XCStrings parser and marshaler

cungminh2710:
💡 What: Optimized `XCStringsParser` and `MarshalXCStrings` by eliminating redundant map key sorting, using `strings.Builder` for key/context generation, and optimizing slice cloning in recursive functions.
🎯 Why: Reducing $O(N \log N)$ overhead and heap allocations in hot paths significantly improves performance for large string catalogs.
📊 Impact: Improves parsing performance by ~3-7% and marshaling performance by ~4-5% for large catalogs.
🔬 Measurement: Verified with `BenchmarkXCStringsParser_ParseLarge` and `BenchmarkMarshalXCStrings_Large` in `internal/i18n/translationfileparser/xcstrings_benchmark_test.go`.

---
*PR created automatically by Jules for task [16082339209001710557](https://jules.google.com/task/16082339209001710557) started by @cungminh2710*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -201,3 +201,7 @@
 ## 2026-10-20 - Optimizing JSONC comment parsing via manual byte scanning
 **Learning:** Replacing line-by-line regular expression matching with a manual `[]byte` scanner and avoiding per-line `string` conversions significantly reduces allocation overhead and CPU time in translation file parsers. Regex overhead, especially for simple key/value patterns, is often a hidden bottleneck compared to optimized byte scanning.
 **Action:** Optimized `parseJSONCKeyComments` and its helper functions in `internal/i18n/translationfileparser/jsonc_parser.go` to operate entirely on `[]byte`, resulting in a ~30% speedup and ~17% fewer allocations.
+
+## 2026-10-25 - Optimizing XCStrings parser and marshaler via sort elimination and string building
+**Learning:** For structured catalog formats like Apple's XCStrings (JSON-based), intermediate O(N log N) sorting of map keys during traversal is often redundant and adds significant CPU/allocation overhead for large files. Direct map iteration is sufficient when the final output is either a map or passed to `json.MarshalIndent` (which sorts keys internally). Additionally, using `strings.Builder` with heuristic `Grow` hints for compound key and multiline context generation further reduces allocation pressure.
+**Action:** Removed redundant sorting and refactored string/slice manipulation in `internal/i18n/translationfileparser/xcstrings_parser.go`.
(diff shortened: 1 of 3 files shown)


cungminh2710:

---

### Issue 1 of 3
internal/i18n/translationfileparser/xcstrings_parser.go:299
This is the one `steps` copy that the PR didn't migrate to the new `make+copy` pattern — it still uses `append([]xcstringsPathStep(nil), steps...)`. Keeping it consistent with the rest of the file is worth it for readability and to ensure the same allocation behaviour the PR targets.

```suggestion
		stepsCopy := make([]xcstringsPathStep, len(steps))
		copy(stepsCopy, steps)
		*refs = append(*refs, xcstringsLeafRef{key: xcstringsKeyForSteps(baseKey, steps), steps: stepsCopy})
```

### Issue 2 of 3
internal/i18n/translationfileparser/xcstrings_benchmark_test.go:64-89
The fixture is assembled with `+=` inside a loop, so each iteration copies the entire string so far — O(N²) work in the setup phase. While it doesn't affect benchmark accuracy (it runs before `b.ResetTimer()`), it makes the benchmark binary itself slower to initialise. Switching to `strings.Builder` here matches the optimization applied in the production code and keeps the benchmark startup fast even if `numEntries` is scaled up.

```suggestion
	numEntries := 500
	var sb strings.Builder
	sb.WriteString("{\n  \"sourceLanguage\": \"en\",\n  \"strings\": {")
	for i := 0; i < numEntries; i++ {
		if i > 0 {
			sb.WriteByte(',')
		}
		fmt.Fprintf(&sb, `
    "key_%d": {
      "comment": "Comment for key %d",
      "localizations": {
        "en": {
          "stringUnit": {
            "state": "translated",
            "value": "Value %d"
          }
        }
      }
    }`, i, i, i)
	}
	sb.WriteString("\n  },\n  \"version\": \"1.0\"\n}")
	contentBytes := []byte(sb.String())
```

### Issue 3 of 3
internal/i18n/translationfileparser/xcstrings_parser.go:63-84
**Shadowed `value` variable in outer loop**

The outer loop introduces `value` via `for key, value := range stringsNode`, which is immediately consumed by `xcstringsObjectValue(value, ...)`. Inside the inner `for _, ref := range refs` block, `value, ok, err := xcstringsValueForRef(...)` re-declares `value` in the same function scope, silently shadowing the outer one. Go accepts this because `:=` creates a new variable in the inner block, so there is no runtime bug, but if a future refactor moves the inner assignment or changes the outer usage, the shadowing becomes a subtle trap. Renaming the inner variable (e.g. `translatedValue`) would eliminate the ambiguity.

PR #1150, opened 2026-06-29
feat(cat): add native project comment support in CAT workspace

cungminh2710:
## Summary

Adds first-class CAT comment support for native projects. Comments and issues are stored in Postgres, surfaced on CAT segments, and writable through the existing comment API.

## Changes

- **Schema**: new `project_translation_comments` table (migration `0054`) with type (`comment`/`issue`), status, locale, author, and optional issue type
- **Service**: `NativeCatCommentService` handles list/save/resolve; `NativeCatService` loads comments onto segments
- **API**: native projects use `POST/PATCH .../cat/comments` instead of returning `provider_cat_unsupported`
- **Queue filter**: `has_issues` now works for native projects via DB-backed unresolved issues
- **UI**: native projects enable comment/issue input (`canAddComments`, issue tabs, has-issues filter)
- **Review fixes**: strict null checks, org-scoped key lookup, single-query author loading, reviewer-gated issue resolution, extracted comment intelligence helper

## Testing

- `vp test src/lib/projects/cat/ src/components/cat/cat-queue-filter.test.ts src/components/cat/project-file-cat-mapper.test.ts src/api/routes/project/project-cat.route.test.ts`
- `vp check --fix`
- `vp run db:migrate`

<div><a href="https://cursor.com/agents/bc-ae37fa22-9f07-46e4-8d10-6231b29f9c83"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ae37fa22-9f07-46e4-8d10-6231b29f9c83"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/hyperlocalise-web/drizzle/0054_mean_speed.sql
+++ b/apps/hyperlocalise-web/drizzle/0054_mean_speed.sql
@@ -1,0 +1,22 @@
+CREATE TYPE "public"."project_translation_comment_type" AS ENUM('comment', 'issue');--> statement-breakpoint
+CREATE TABLE "project_translation_comments" (
+	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+	"organization_id" uuid NOT NULL,
+	"project_id" text NOT NULL,
+	"translation_key_id" uuid NOT NULL,
+	"target_locale" text NOT NULL,
+	"type" "project_translation_comment_type" DEFAULT 'comment' NOT NULL,
+	"status" text,
+	"text" text NOT NULL,
+	"issue_type" text,
+	"author_user_id" uuid,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"updated_at" timestamp with time zone DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+ALTER TABLE "project_translation_comments" ADD CONSTRAINT "project_translation_comments_organization_id_organizations_id_fk" FOREIGN KEY ("organization_id") REFERENCES "public"."organizations"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "project_translation_comments" ADD CONSTRAINT "project_translation_comments_project_id_projects_id_fk" FOREIGN KEY ("project_id") REFERENCES "public"."projects"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "project_translation_comments" ADD CONSTRAINT "project_translation_comments_translation_key_id_project_translation_keys_id_fk" FOREIGN KEY ("translation_key_id") REFERENCES "public"."project_translation_keys"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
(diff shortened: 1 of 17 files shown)


cungminh2710:

---

### Issue 1 of 5
apps/hyperlocalise-web/src/components/cat/cat-queue-filter.ts:25
**Loose equality for null check**

Line 25 uses `==` instead of `===` to compare `providerKind` with `null`. While the loose-equality check intentionally matches both `null` and `undefined`, this differs from the strict equality style used in the same condition (`providerKind === "crowdin"`). If `providerKind` can genuinely be `undefined`, document that intent; otherwise use `=== null` to be consistent with the surrounding code.

```suggestion
    return providerKind === "crowdin" || providerKind === "phrase" || providerKind === null;
```

### Issue 2 of 5
apps/hyperlocalise-web/src/lib/projects/cat/native-cat-comment-service.ts:113-130
**`save` does not verify cross-org ownership of `translationKeyId`**

The `save` method looks up the key by matching `id`, `projectId`, and `repositorySourceFileId`. However, `input.translationKeyId` (the caller-supplied `externalStringId` from the POST body) is accepted at face value and passed straight to the `WHERE id = ?` clause. If `translationKeyId` belongs to a different project or organization, the join on `projectId` and `sourceFile.id` should prevent a write — but the `sourceFile` lookup only checks `organizationId` + `projectId` + `sourcePath`, so a source-path collision across projects is the only guard against a cross-project injection. Explicitly adding `eq(schema.projectTranslationKeys.organizationId, input.organizationId)` — if the column exists — or at minimum confirming via the `organizationId` on the fetched key would harden this boundary.

### Issue 3 of 5
apps/hyperlocalise-web/src/lib/projects/cat/native-cat-comment-service.ts:156-175
**Two separate DB round-trips after insert instead of one JOIN**

After inserting a comment, the code issues a second `SELECT` to fetch the author's name/email for the returned `ProjectFileCatComment`. These can be combined into a single query by joining `users` inside the `RETURNING`-based SELECT (or by fetching the user data before the insert). At high throughput this doubles the DB load and leaves a tiny window where the user row could be deleted between the two queries. The `resolve` method has the same pattern.

### Issue 4 of 5
apps/hyperlocalise-web/src/lib/projects/cat/native-cat-comment-service.ts:200-230
**`resolve` allows any authenticated project member to resolve another user's issue**

The `resolve` method only verifies that the comment belongs to the correct organization and project — it does not check whether `actorUserId` is the original author or holds an elevated role. A translator could resolve an issue opened by a reviewer on any key in the same project. If role-based resolution gating is a requirement (e.g., only the opener or a reviewer may resolve), this needs an explicit check before the UPDATE.

### Issue 5 of 5
apps/hyperlocalise-web/src/components/cat/project-file-cat-mapper.ts:155-156
**Long ternary chains degrade readability**

Both `productMeaning` expressions in `intelligenceFor` and `segmentIntelligenceFor` have grown into multi-condition inline ternaries that are hard to reason about and easy to get wrong during future edits. Extracting a small helper such as `buildCommentProductMeaning(commentCount, issueCount, isNativeProject)` would make the intent clearer and avoid duplication of the issue-count string-building logic between the two functions.

PR #1195, opened 2026-07-03
feat(web): split CAT queue API and polish skeleton loading

cungminh2710:
## What changed

Split the CAT workspace into per-section API calls and improved loading UX:

- **API:** Added `GET .../cat/queue/summary` and `GET .../cat/queue`; kept legacy `GET .../cat` for compatibility. Crowdin queue builds skip summary on page fetches and scope approvals/issues to the visible page.
- **Client:** Separate React Query hooks for queue summary, queue page, and segment detail; skeleton list replaces stale rows during filter/search/pagination refetch.
- **Layout:** Compact job CAT toolbar; each panel scrolls independently within a fixed viewport height.
- **Skeleton polish:** Full 3-panel shell on cold start, summary/progress skeletons while counts load, comment skeletons on segment detail fetch.

## How to test

- Open a Crowdin job file in CAT (`/jobs/.../strings?sourcePath=...`).
- Confirm cold start shows the 3-panel skeleton (not a centered spinner).
- Change queue filter/search and confirm the list shows skeleton rows instead of stale segments.
- Select a segment and confirm comments skeleton briefly while detail loads.
- Run `cd apps/hyperlocalise-web && vp check && vp test src/components/cat/project-file-cat-api.test.ts src/lib/providers/tms-provider-live-cat.test.ts`.

## Checklist

### DIFF
--- a/apps/hyperlocalise-web/src/api/routes/project/project.route.ts
+++ b/apps/hyperlocalise-web/src/api/routes/project/project.route.ts
@@ -29,6 +29,7 @@ import { sourceContentType } from "@/lib/file-storage/source-file-metadata";
 import {
   countTmsProviderLiveOpenJobsForProject,
   getTmsProviderLiveCatFile,
+  getTmsProviderLiveCatQueueSummary,
   getTmsProviderLiveCatSegmentDetail,
   getTmsProviderLiveFileDetail,
   getTmsProviderLiveProject,
@@ -40,13 +41,17 @@ import {
 import { listOrganizationProjects } from "@/lib/projects/organization/organization-project-service";
 import {
   getNativeProjectCatFile,
+  getNativeProjectCatQueueSummary,
   getNativeProjectCatSegmentDetail,
   resolveNativeProjectCatComment,
   saveNativeProjectCatComment,
   saveNativeProjectCatTranslation,
   updateNativeProjectTranslationStatus,
 } from "@/lib/projects/cat/native-cat-service";
-import { resolveProjectFileCatPagination } from "@/lib/projects/cat/project-file-cat-pagination";
+import {
+  resolveProjectFileCatIncludeQueueSummary,
+  resolveProjectFileCatPagination,
+} from "@/lib/projects/cat/project-file-cat-pagination";
 import {
   getProjectFileDetail,
   listFilteredProjectFiles,
@@ -74,6 +79,7 @@ import {
   projectFileCatSegmentParamsSchema,
   projectFileCatSegmentQuerySchema,
   projectFileCatQuerySchema,
+  projectFileCatQueueSummaryQuerySchema,
   projectFileCatConcordanceBodySchema,
   projectFileCatCommentBodySchema,
   projectFileCatCommentResolveBodySchema,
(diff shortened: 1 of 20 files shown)


cungminh2710:
@cursoragent 
---

### Issue 1 of 4
apps/hyperlocalise-web/src/lib/providers/tms-provider-live-cat.test.ts:1054
**Weakened assertion may hide issue-detection regression**

The assertion was relaxed from `>= 2` to `>= 1` to accommodate the new page-scoped `listStringCommentsForStrings` approach replacing the file-wide CROQL scan. If the `has_issues` CROQL filter still selects the same 2+ strings but `loadCrowdinVisibleStringIdsWithUnresolvedIssues` only maps 1 of them as having issues (e.g. because `listStringCommentsForStrings` with `targetLanguageId` applies stricter matching than the CROQL filter), the effective filter silently mislabels segments, and this test would still pass.

### Issue 2 of 4
apps/hyperlocalise-web/src/lib/providers/tms-provider-live.ts:1981-1995
**Phrase error may fall through to Crowdin summary path**

If `mapPhraseLiveCatError(error)` does not always rethrow, execution falls through to `getCrowdinLiveCatQueueSummary` with a Phrase project context, producing incorrect data or a runtime Crowdin API error for Phrase users. The analogous `getTmsProviderLiveCatFile` uses `return await buildPhraseLiveCatFile(...)` in the try block and `mapPhraseLiveCatError` in the catch, relying on the same implicit-rethrow contract — but there the fallthrough would be to `buildCrowdinLiveCatFile`, which has historically been safe. For the new summary path the fallthrough target is different code, so a defensive `return` or explicit `throw` after `mapPhraseLiveCatError(error)` would make the intent unambiguous.

### Issue 3 of 4
apps/hyperlocalise-web/src/lib/providers/tms-provider-live.ts:1961
**Phrase path returns `queueSummary` that is now typed `undefined`**

`buildPhraseLiveCatFile` is called without `includeQueueSummary`, so the summary is computed. However, `catFile.queueSummary` is now `ProjectFileCatQueueSummary | undefined` because the schema changed it to optional. TypeScript therefore does not guarantee the value is present. In the route handler `if (!queueSummary) return projectNotFoundResponse(c)` would fire for a Phrase project if the field were ever absent, returning 404 even though the file was found. Consider asserting the value is defined before returning, or updating the return type to reflect the invariant that this code path always produces a summary.

### Issue 4 of 4
apps/hyperlocalise-web/src/components/cat/project-file-cat-workspace.tsx:1394-1396
**`isQueueLoading` only covers `isPlaceholderData` — misses direct re-fetches**

`catQuery.isPlaceholderData` is `true` only when the query was seeded from `placeholderData: keepPreviousData`. A direct pagination click or filter change that bypasses placeholder data (e.g. when `keepPreviousData` is not configured for `useCatSegmentQuery`, or on the very first fetch after mounting with a new key) would leave `isPlaceholderData === false` while `catQuery.isFetching` is `true`, making `isQueueLoading` `false` and skipping the skeleton. Confirm `useCatSegmentQuery` uses `placeholderData: keepPreviousData`; otherwise the skeleton will not appear during pagination re-fetches.

PR #1227, opened 2026-07-04
fix(cat): keep editor drafts out of queue pagination hydration

cungminh2710:
## What changed

Queue pagination no longer writes translation state into editor drafts. Hydration now updates `segmentMeta` only; `targetText` and `status` live in `CatSegmentDraft` (lazy target sync, user edits, approve/save).

Also included:
- Fix concordance panel stuck in loading when a newer review sequence supersedes the lookup
- Skip concordance re-fetch once loaded (including AI review)
- Remove fake `targetText`/`status` synthesis from queue snapshots
- Delete unused legacy `cat-dirty-state` helpers (MobX drafts own dirty state now)

## How to test

- Open a project-file CAT workspace, select a segment, confirm translation loads in the editor
- Click **Load more** in the queue — selected segment translation should remain visible
- Open intelligence panel — concordance loads once; trigger AI review — glossary/TM should not re-fetch or flash skeleton
- Run `cd apps/hyperlocalise-web && vp test src/components/cat/workspace/`

## Checklist


cungminh2710:

---

### Issue 1 of 2
apps/hyperlocalise-web/src/components/cat/workspace/store/cat-workspace-store.ts:450-455
**Skipped segments lose their status on initial hydration**

The new guard `if (segment.targetText.trim())` skips draft creation for any segment with an empty or whitespace-only `targetText`. Segments with `status: "skipped"` (or `"needs_review"`) and an empty translation — a common legitimate state — will have no draft, so `getQueuePanelSegments` and `matchesQueueFilter` both fall back to `"pending"`. A user who applies a queue filter to hide skipped segments will see those segments incorrectly reappear as pending until they explicitly load each one via lazy target sync. The condition should also pass through segments whose status is not `"pending"`, regardless of target content, to preserve filter correctness.

### Issue 2 of 2
apps/hyperlocalise-web/src/components/cat/workspace/use-cat-workspace-controller.ts:224-225
When `onReviewWithAi` fires while the intelligence panel's concordance fetch is still in flight, `concordanceLoadedSegmentIdsRef` doesn't contain the segment yet so `shouldLookupConcordance` is `true` in both calls. This causes two real network requests for the same segment. The second request's result is used; the first is discarded by the stale-sequence guard. Checking `store.concordanceLoadingSegmentId` avoids the redundant call.

```suggestion
      const shouldLookupConcordance =
        includeConcordance &&
        !concordanceLoadedSegmentIdsRef.current.has(segmentId) &&
        store.concordanceLoadingSegmentId !== segmentId;
```

PR #1230, opened 2026-07-04
feat(cat-validate): add segment validation service and profile rules

cungminh2710:
## What changed

Extracted shared segment validation into `internal/i18n/segmentvalidate`, refactored `runsvc` to use it, and added a new `apps/cat-validate` HTTP service for CAT segment checks.

The service exposes `GET /health` and `POST /v1/validate/segment` with WorkOS JWT auth. Deploy config moves to repo-root `vercel.json` with a dedicated `cat_validate` service and rewrite to `/api/cat-validate/:path*`.

Always-on Crowdin-inspired profile rules now run after format validation:
- Extra placeholders ICU misses (`%s`, `%(name)s`, `${var}`, etc.)
- Whitespace profile (leading/trailing spaces, tabs, NBSP; `&nbsp;` → `\u00A0`)
- Special char literal parity (`\n`, `\t`, `\uXXXX`, `\xHH`, etc.)

Optional QA modes remain opt-in via `modes`: `same_as_source`, `whitespace_only`, `not_localized`.

## How to test

- `make fmt`
- `make lint`
- `make test`
- `bazel build //apps/cat-validate:cat_validate`
- `go test ./internal/i18n/segmentvalidate/... ./apps/cat-validate/...`

Manual API check (with valid WorkOS JWT):

```bash
curl -X POST http://localhost:$PORT/v1/validate/segment \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"sourceText":" Hello %s ","targetText":" Bonjour %s ","sourcePath":"/messages/en.json"}'
```

## Checklist


cungminh2710:
@cursoragent update pr description to use session 

Fix the following 5 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 5
apps/cat-validate/auth.go:224-243
**Auth accepts cookie only, but PR curl example sends Bearer header**

The middleware reads exclusively from the `wos-session` cookie (`sessionCookieValue`), but the PR description documents a curl invocation that sends `Authorization: Bearer $TOKEN`. Any caller following that example will receive `401 Unauthorized` because the header is never inspected. The service is usable from browser contexts where the cookie is set by AuthKit, but machine-to-machine or CLI clients cannot authenticate with a static token. Consider either (a) also extracting a bearer JWT from the `Authorization` header as a fallback, or (b) updating the PR description to remove the misleading Bearer example.

### Issue 2 of 5
internal/i18n/segmentvalidate/validate.go:39
**MaxLength check uses byte length, not rune/character count**

`len(req.TargetText)` returns the number of UTF-8 bytes, not characters. A single CJK or Arabic character is 2–4 bytes, so a 10-character Chinese translation of a 10-character English string will report an excess length. This is especially harmful for a localisation tool where translated text often uses multi-byte scripts.

```suggestion
	if req.MaxLength > 0 && len([]rune(req.TargetText)) > req.MaxLength {
```

### Issue 3 of 5
internal/i18n/segmentvalidate/validate.go:79-115
**Profile parity silently extends to Markdown segments (breaking behaviour change for the CLI)**

The old `runsvc` markdown case returned `nil` after its own checks; the new `validateForKind` falls through to `validateProfileParity` for every format kind, including `FormatMarkdown`. This means that in the CLI today, markdown translation outputs are now also subject to whitespace-profile (leading/trailing space, NBSP count) and special-char-parity (`\n`, `\t`, `\uXXXX`) rules that were never applied before. A markdown source that intentionally uses `\n` literals or has trailing whitespace will now fail post-translate validation and block the CLI, while it passed previously. The `TestValidateTranslatedOutputMatrix` cases that exercised these markdown paths were removed, hiding the regression.

### Issue 4 of 5
internal/i18n/segmentvalidate/extra_placeholders.go:407-411
**`%i` is matched but is not a recognised format specifier in most runtimes**

The pattern `%[sidf@]\b` includes `%i`, which is not a standard format specifier in Python, Ruby, or Swift (common consumers of `.json` i18n files). This can produce false-positive placeholder-parity failures when a source uses `%i` in ordinary prose (e.g. "step %i of %d") and a translator's output legitimately omits or rearranges it in a language where that context reads differently. Consider restricting the character class to the specifiers actually used in your projects, or document the intended superset explicitly.

PR #1235, opened 2026-07-04
🧪 Scout: add TestParseASTSelectOrdinalWithOffset

cungminh2710:
💡 What: Added a unit test `TestParseASTSelectOrdinalWithOffset` to `internal/i18n/icuparser/parse_test.go`.
🎯 Why: Ensures that `selectordinal` blocks correctly support the `offset` parameter, matching the behavior of standard `plural` blocks and preventing regressions in complex ICU message parsing.
🧩 Scope: `internal/i18n/icuparser/parse_test.go`.
✅ Verification: Ran `go test -v ./internal/i18n/icuparser/...` to verify the new test passes and ensure no regressions were introduced.
🛡️ Confidence: High. The test follows existing patterns and specifically targets a shared code path for plural/ordinal options.

---
*PR created automatically by Jules for task [16043799984554746567](https://jules.google.com/task/16043799984554746567) started by @cungminh2710*

### DIFF
--- a/internal/i18n/icuparser/parse_test.go
+++ b/internal/i18n/icuparser/parse_test.go
@@ -540,27 +540,93 @@ func TestParseASTSelect(t *testing.T) {
 
 func TestParseASTSelectOrdinalWithOffset(t *testing.T) {
 	// selectordinal blocks should support the offset parameter just like plural blocks.
-	msg := "{count, selectordinal, offset:1 one {# item} other {# items}}"
-	elems, err := Parse(msg, nil)
-	if err != nil {
-		t.Fatalf("Parse failed: %v", err)
-	}
-	if len(elems) != 1 {
-		t.Fatalf("expected 1 element, got %d", len(elems))
+	tests := []struct {
+		name           string
+		msg            string
+		expectedOffset int
+	}{
+		{
+			name:           "no space",
+			msg:            "{count, selectordinal, offset:1 one {# item} other {# items}}",
+			expectedOffset: 1,
+		},
+		{
+			name:           "space after colon",
+			msg:            "{count, selectordinal, offset: 1 one {# item} other {# items}}",
+			expectedOffset: 1,
+		},
+		{
+			name:           "space before colon",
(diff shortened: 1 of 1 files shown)


cungminh2710:

---

### Issue 1 of 3
internal/i18n/icuparser/parse_test.go:562-564
**Missing option content validation**

The test confirms there are 2 options but never inspects their selectors (`"one"`, `"other"`) or their parsed values. The comparable `TestParseASTSelect` verifies each selector and its literal content. Without this, the test would pass even if the options were parsed with wrong selectors or garbled values — reducing its regression value for the exact scenario the PR description calls out.

### Issue 2 of 3
internal/i18n/icuparser/parse_test.go:556-561
`pl.Value` and `pl.Ordinal` are unverified

`PluralElement` carries both a `Value` field (the argument name, `"count"`) and a separate boolean `Ordinal` field that the parser sets independently of `PluralType`. The test checks `pl.Type()` but leaves `pl.Value` and `pl.Ordinal` unexamined. If either were wrong the test would still pass.

```suggestion
	if pl.Value != "count" {
		t.Errorf("expected argument %q, got %q", "count", pl.Value)
	}
	if pl.Type() != TypeSelectOrdinal {
		t.Errorf("expected type %q, got %q", TypeSelectOrdinal, pl.Type())
	}
	if !pl.Ordinal {
		t.Errorf("expected Ordinal to be true for selectordinal")
	}
	if pl.Offset != 1 {
		t.Errorf("expected offset 1, got %d", pl.Offset)
	}
```

### Issue 3 of 3
internal/i18n/icuparser/parse_test.go:541-543
**Single-case test leaves offset-format variants uncovered for `selectordinal`**

`TestParsePluralWithSpaceInOffset` exercises five offset syntax variants (no space, space after colon, space before colon, multiple spaces, negative value) for `plural`. The new test covers only `offset:1` for `selectordinal`. While both types share `parsePluralOptions`, a table-driven approach mirroring the existing test would catch any future divergence and would match the testing pattern established for this exact feature.

PR #1245, opened 2026-07-05
fix(web): auto-open first job file in CAT workspace

cungminh2710:
## What changed

Job strings pages no longer stop on a source file picker when a file can be opened automatically. The first openable file is selected and the user is redirected straight into the CAT workspace, with a brief loading state instead of a broken click-to-select step.

This fixes a bug where the first file appeared pre-selected in the tree but clicking it did nothing, because navigation only ran on selection change.

## How to test

1. Open a Crowdin-backed job strings page without `sourcePath` in the URL, e.g. `/org/{org}/projects/{project}/jobs/{job}/strings?targetLocale=vi`.
2. Confirm the page briefly shows "Opening workspace…" and then loads the CAT editor for the first source file.
3. Confirm the task detail page file list still works with "View strings" and double-click when `openInCatOnSelect` is off.
4. Run `vp test src/app/[lang]/(authenticated)/org/[organizationSlug]/projects/[projectId]/jobs/[jobId]/_components/tms/job-source-files-panel.test.tsx`.

## Checklist


cungminh2710:

---

### Issue 1 of 2
apps/hyperlocalise-web/src/app/[lang]/(authenticated)/org/[organizationSlug]/projects/[projectId]/jobs/[jobId]/strings/_components/load-job-cat-files.ts:254-272
**Double `fetchJobDetail` for native jobs**

For a native (non-provider) job, `loadJobCatJobSourceFiles` fetches the job detail endpoint twice: once inside `loadJobCatProviderJobFiles` (which calls `fetchJobDetail` when `parseProviderJobId` returns null, checks project membership, then returns `[]` because there is no `externalProviderKind`), and a second time at line 265 in this function's fallback path. Each page load with no file reference incurs two sequential round-trips to the same endpoint instead of one.

### Issue 2 of 2
apps/hyperlocalise-web/src/app/[lang]/(authenticated)/org/[organizationSlug]/projects/[projectId]/jobs/[jobId]/strings/_components/job-cat-page-content.tsx:297-307
**Misleading empty-state message when `targetLocale` is null**

`resolveDefaultJobCatFileReference` returns `null` whenever `requestedTargetLocale` is `null` for a native file (because `resolveTargetLocale` passes it through directly, then the `!targetLocale` guard skips the file). If the user lands on the strings page without a `targetLocale` query param but there is a valid source file, this branch renders "No source file is linked to this task." — which is factually wrong. A message like "No target locale is specified for this task." would be more accurate in that case.

PR #1247, opened 2026-07-05
feat(web): add projects UX utilities for avatars and recent tracking

cungminh2710:
## What changed

This draft PR adds the foundational utilities for the projects page UX refresh:

- `ProjectAvatar` component for TMS/native project logos with provider badge overlay
- `recent-projects` helpers to persist and resolve recently opened projects per organization
- `sanitizeCrowdinProjectLogo` for safe Crowdin logo URLs (data URLs and https)

Follow-up commits will wire these into the projects page layout, Crowdin metadata mapping, and TMS sorting.

## How to test

- Run `vp test src/app/[lang]/\(authenticated\)/org/[organizationSlug]/projects/_components/recent-projects.test.ts`
- Run `vp test src/lib/providers/adapters/crowdin/crowdin-project-logo.test.ts`
- Verify `ProjectAvatar` renders initials fallback when `logoUrl` is absent

## Checklist

### DIFF
--- a/apps/hyperlocalise-web/src/app/[lang]/(authenticated)/org/[organizationSlug]/projects/[projectId]/_components/project-overview.fixture.ts
+++ b/apps/hyperlocalise-web/src/app/[lang]/(authenticated)/org/[organizationSlug]/projects/[projectId]/_components/project-overview.fixture.ts
@@ -20,6 +20,8 @@ export const projectOverviewFixture: ProjectListRow = {
   targetLocales: ["fr-FR", "de-DE", "es-ES"],
   externalProjectUrl: "https://crowdin.com/project/website",
   isActive: true,
+  logoUrl: null,
+  lastActivityAt: new Date(Date.now() - 2 * 60 * 60 * 1000).toISOString(),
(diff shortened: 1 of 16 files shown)


cungminh2710:

---

### Issue 1 of 1
apps/hyperlocalise-web/src/app/[lang]/(authenticated)/org/[organizationSlug]/projects/_components/projects-page-content.tsx:263
**`hasFilteredResults` ignores the active `sourceFilter`**

`hasFilteredResults` always ORs both `filteredNativeProjects` and `filteredTmsProjects` regardless of which tab is visible. When `sourceFilter === "tms"` and a search query returns no TMS matches but does return native matches, `hasFilteredResults` is `true` because of the hidden native results — so the "No projects match your search" banner is suppressed. The user sees the TMS section render its "No TMS projects found" empty state (which reads like an API failure, not a search miss) with no corrective message. The same blind-spot exists in the `"native"` tab when TMS has matches.

The flag should only consider whichever sections are actually visible:

```ts
const hasFilteredResults =
  (showNativeSection && filteredNativeProjects.length > 0) ||
  (showTmsSection && filteredTmsProjects.length > 0);
```

PR #1249, opened 2026-07-05
feat(web): select GitHub repo for chat context

cungminh2710:
## What changed

- Replaced chat and inbox composer project/context controls with enabled GitHub repository selection.
- Auto-selects a single enabled repository and sends `repositoryFullName` with chat create/reply requests.
- Validates selected repositories server-side and seeds the web conversation repository session for Hyperlocalise agent repo tools.

## How to test

- `vp check --fix`
- `vp test src/api/routes/conversation/conversation.route.test.ts`
- `vp test src/agents/hyperlocalise/agent/channels/web.test.ts`
- `vp test` currently still fails in unrelated public API/MCP auth tests with 403/401/400 status mismatches.

## Checklist

### DIFF
--- a/apps/hyperlocalise-web/src/api/routes/conversation/conversation.route.test.ts
+++ b/apps/hyperlocalise-web/src/api/routes/conversation/conversation.route.test.ts
@@ -1,5 +1,7 @@
 import "dotenv/config";
 
+import { randomInt } from "node:crypto";
+
 import { eq } from "drizzle-orm";
 import { testClient } from "hono/testing";
 import { afterEach, beforeAll, describe, expect, it, vi } from "vite-plus/test";
@@ -7,6 +9,7 @@ import { afterEach, beforeAll, describe, expect, it, vi } from "vite-plus/test";
 import { createApp } from "@/api/app";
 import { db, schema } from "@/lib/database";
 import type { FileStorageAdapter } from "@/lib/file-storage";
+import { getWebConversationRepositorySession } from "@/lib/agent-runtime/loops/conversation-repository-session";
 import { createMemoryFileStorageAdapter } from "../file/file.fixture";
 import { createProjectTestFixture } from "../project/project.fixture";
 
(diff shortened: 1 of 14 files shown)


cungminh2710:

---

### Issue 1 of 3
apps/hyperlocalise-web/src/api/routes/conversation/conversation.route.ts:464
Missing validation for `repositoryFullName` in the reply path. The create-conversation handler parses this field through `createConversationRequestSchema` which applies `.trim()` and `.max(255)`, but the reply handler reads it directly with `asString()`. An arbitrarily large or untrimmed value is passed straight to `resolveWebProjectRepositoryGitHubContext` — a trailing space would cause a silent lookup miss, and there is no length ceiling before the DB query.

```suggestion
        const rawRepositoryFullName = asString(body.repositoryFullName);
        const repositoryFullName = rawRepositoryFullName
          ? rawRepositoryFullName.trim().slice(0, 255) || undefined
          : undefined;
```

### Issue 2 of 3
apps/hyperlocalise-web/src/api/routes/conversation/conversation.route.ts:562-564
**Throw after message is committed may leave attachments orphaned.** `seedConversationRepositorySession` is called inside the `try` block after `addInteractionMessage` has already persisted the message and its file references. If all three CAS retries fail, the `catch` on line 565 runs `cleanupStoredFiles`, deleting the blob-storage objects and their DB records — but the message row (which references those file URLs) remains. Users would see a stored message whose attachments 404. Moving the seeding call before `addInteractionMessage`, or catching the seeding error separately without running file cleanup, would avoid this inconsistency.

### Issue 3 of 3
apps/hyperlocalise-web/src/app/[lang]/(authenticated)/org/[organizationSlug]/chat/_components/chat-page-content.tsx:75-164
**Duplicate `RepositorySelector` component and repository type.** An almost identical `RepositorySelector` (and the accompanying `ChatGithubRepository` type) is defined in this file and again in `reply-composer.tsx`. The only structural difference is the button primitive (`Button` here vs `PromptInputButton` there). Diverging the two implementations over time is an easy source of bugs. Consider extracting the shared component and the common repository type (`InboxGithubRepository` already lives in `inbox-api.ts`) into a collocated shared file so both composers can import from the same place.

PR #1422, opened 2026-07-15
fix(chat): share one stream owner between inbox and dock

cungminh2710:
## What changed

Inbox and chat dock each owned a client stream runner, so opening the same conversation in both could start two `/chat` requests and diverge UI.

`ChatStreamManager` is now the only client stream owner. Snapshots live on `ChatDockStore` by conversation id (not only dock tabs), inbox attaches without creating tabs, `start()` is a no-op when already streaming, and closing a dock tab no longer aborts an in-flight stream.

## How to test

- [ ] Start a chat in the dock, then Open in Inbox while tools are still streaming — only one agent turn should run, and both surfaces show the same live parts
- [ ] Start a chat from inbox only — no dock tab should appear; stream still works
- [ ] Close the dock tab mid-stream while inbox is open — stream should continue in inbox
- [ ] `vp test src/components/app-shell/chat-dock/chat-dock-store.test.ts src/components/app-shell/chat-dock/chat-stream-manager.test.ts`
- [ ] `vp check --fix` in `apps/hyperlocalise-web`

## Checklist

- [x] I ran relevant checks locally (`vp test` for chat-dock tests, `vp check --fix`)


cungminh2710:

---

### Issue 1 of 5
apps/hyperlocalise-web/src/components/app-shell/chat-dock/chat-stream-manager.ts:169-178
**Stale store silently used after first creation**

`getChatStreamManager` caches the manager by `organizationSlug` and discards the `store` argument on every call after the first. Both `InboxPageContent` and `ChatDockBridge` now call this factory on every render, each passing `chatDock` from the same `useAppShellStore()` singleton, so no bug fires today. However, if a test, a story, or a future call-site constructs a fresh `ChatDockStore` and passes it in — expecting to get a manager wired to that store — it will silently receive the original manager pointing at the first store. Consider accepting the store only in a constructor/factory that is called once, or validating that the passed store matches the cached instance and logging/throwing if not.

### Issue 2 of 5
apps/hyperlocalise-web/src/components/app-shell/chat-dock/chat-stream-manager.ts:140-161
**Aborted streams leave a stale "complete" snapshot in the store**

When `stop()` is called it synchronously (1) calls `controller.abort()`, (2) removes the stream from `activeStreams`, and (3) calls `store.markStreaming(conversationId, false)`, which transitions the snapshot in `streamsByConversationId` to `status: "complete"`. The resulting `AbortError` is caught by the early-return branch in `start()`, so the `finally` block runs but `finishedSuccessfully` is false — meaning `onStreamFinished` is never invoked. Because `onStreamFinished` is the only code path that calls `chatDock.clearStreamSnapshot(conversationId)`, the "complete" entry sits in `streamsByConversationId` indefinitely and will be returned by `getStreamSnapshot()` as a stale `streamedAssistant` until the next `hydrate()` resets the map. The practical impact is limited to org-navigation scenarios (the only current caller of `stop()`), since `hydrate()` follows shortly after. But if future callers invoke `stop()` independently, the stale snapshot will be visible to the inbox and the dock panel.

### Issue 3 of 5
apps/hyperlocalise-web/src/components/app-shell/chat-dock/chat-dock-panel.tsx:280-284
**Repeated `isStreaming` expression used three times**

The compound check `streamManager.isStreaming(tab.id) || store.getStreamSnapshot(tab.id)?.status === "streaming"` is inlined three separate times (lines 360–361, 373–374, and in `isBusy` slightly differently via `tab.isStreaming`). Extracting it once makes the intent clear and avoids future divergence.

```suggestion
  const isTabStreaming =
    streamManager.isStreaming(tab.id) ||
    store.getStreamSnapshot(tab.id)?.status === "streaming";
  const isBusy =
    createConversationMutation.isPending ||
    sendMessageMutation.isPending ||
    isTabStreaming;
```

### Issue 4 of 5
apps/hyperlocalise-web/src/components/app-shell/chat-dock/chat-dock-panel.tsx:266-278
**`?? tab.streamSnapshot` fallback is unreachable**

`clearStreamSnapshot` removes the entry from `streamsByConversationId` and nulls `tab.streamSnapshot` inside the same MobX action transaction. Because both mutations are committed atomically before any reaction fires, `store.getStreamSnapshot(tab.id)` and `tab.streamSnapshot` are always in the same state: if the store entry is null, so is the tab field. The fallback therefore never activates and quietly hides any future divergence.

```suggestion
  const streamedAssistant: StreamedAssistantMessage | null = (() => {
    const snapshot = store.getStreamSnapshot(tab.id);
    if (!snapshot) {
      return null;
    }

    return {
      conversationId: snapshot.conversationId,
      responseToMessageId: snapshot.responseToMessageId,
      message: snapshot.message,
      status: snapshot.status,
    };
  })();
```

### Issue 5 of 5
apps/hyperlocalise-web/src/components/app-shell/chat-dock/chat-dock-store.ts:343-345
**`markStreaming(id, true)` with no prior snapshot is a silent no-op on the central map**

When `isStreaming` is `true` but `current` (the snapshot from `streamsByConversationId`) is null, neither branch of `markStreaming` writes to `streamsByConversationId`. The call only falls through to `updateTab`, which also silently does nothing if the tab doesn't exist. This means calling `markStreaming(id, true)` without a prior `setStreamSnapshot` produces no observable effect on `streamingCount` or `canStartStream`. The current codebase only calls `markStreaming(false)` via `stop()`, so this path never fires today, but the method's name implies it can be used to mark a stream as active. A guard or a comment clarifying that `markStreaming(true)` is a no-op when no snapshot exists would prevent misuse.

```suggestion
  /** Only call with `isStreaming = false` (used by `stop()`). Starting a stream must go through `setStreamSnapshot`. */
  markStreaming(conversationId: string, isStreaming: boolean) {
    const current = this.getStreamSnapshot(conversationId);
    if (!isStreaming && current?.status === "streaming") {
```

PR #1434, opened 2026-07-15
🧪 Scout: improve printf placeholder detection

cungminh2710:
Improved printf-style placeholder detection in `extractExtraPlaceholders` to support a broader range of standard specifiers and modifiers. This ensures that placeholders like `%i`, `%x`, `%u`, or those with width/precision (e.g., `%02d`, `%.2f`) are correctly identified and their presence is enforced during translation validation.

💡 What: Improved printf-style placeholder detection in `extractExtraPlaceholders` to support a broader range of standard specifiers and modifiers.
🎯 Why: The previous implementation missed common specifiers (like `%i`, `%x`, `%u`) and modifiers (width, precision, flags), allowing critical placeholders to be omitted from translations without detection.
🧩 Scope: `internal/i18n/segmentvalidate/extra_placeholders.go` and associated tests.
✅ Verification: Added comprehensive unit tests in `printf_scout_test.go` covering various printf formats. Verified all tests pass in `internal/i18n/segmentvalidate`, `internal/i18n/icuparser`, and `internal/i18n/translationfileparser`.
🛡️ Confidence: Ensures structural parity is strictly enforced for a wider variety of printf-based localization formats, preventing common translation errors.

---
*PR created automatically by Jules for task [8894716972141907948](https://jules.google.com/task/8894716972141907948) started by @cungminh2710*

### DIFF
--- a/.jules/scout.md
+++ b/.jules/scout.md
@@ -143,3 +143,7 @@
 ## 2025-07-20 - [QA Mode and Profile Parity Interaction]
 **Learning:** In `ValidateSegment`, QA checks (like `whitespace_only`) are executed after core format and profile parity checks. If a target string contains only whitespace while the source contains text, `validateWhitespaceProfile` will trigger a `format-whitespace-profile` failure (StatusFail) before the `qa-whitespace-only` warning (StatusWarn) is even evaluated.
 **Action:** When testing QA modes via the top-level `ValidateSegment` entry point, ensure the source and target strings are chosen to either specifically trigger or specifically avoid overlapping profile parity violations to maintain test determinism.
+
+## 2026-07-20 - [Comprehensive Printf Placeholder Detection]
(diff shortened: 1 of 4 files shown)


cungminh2710:

---

### Issue 1 of 2
internal/i18n/segmentvalidate/extra_placeholders.go:14-17
**`%c` / `%C` specifier not included**

The new specifier class `[diuXxfsSFeEgG]` covers a broad set but omits `c` (character) and `C` (wide character, Windows/Swift). `%c` is a standard C printf specifier and could appear in some localization format strings; any such string would silently pass parity validation with the placeholder absent from the translation.

### Issue 2 of 2
internal/i18n/segmentvalidate/extra_placeholders.go:17
**`%@` now accepts width/precision modifiers that are non-standard**

The fourth alternation `%[-+ #0]*[0-9]*(?:\.[0-9]*)?(?:ll|l|hh|h)?@` allows flag and width modifiers before `@`, but `%@` (Objective-C/Swift object description) doesn't support width, precision, or length modifiers in practice. In the unlikely event a string contains something like `%10@` as text (not a real format specifier), this pattern would incorrectly classify it as a placeholder and enforce its presence in translations. Restricting this back to a simple `%@` literal would be safer and matches prior behaviour.

PR #1453, opened 2026-07-17
feat(web): localize workspace teams UI with react-intl

cungminh2710:
## What changed

Localize hardcoded user-facing copy in the workspace teams `_components` folder with react-intl (ICU):

- Colocated `*.messages.ts` modules for list/detail pages, dialogs, form validation, role labels, and toasts
- Client components use `<FormattedMessage>` / `useIntl().formatMessage`
- `team-form` and role helpers accept optional `intl` (same pattern as projects)
- Rich-text messages for Members / invite links

## How to test

- [ ] Open `/org/{slug}/teams` — headers, empty/loading/error states, create/edit/delete dialogs, toasts
- [ ] Open a team detail page — members table, add/remove member, role labels, edit team
- [ ] `cd apps/hyperlocalise-web && vp lint -- src/app/[lang]/(authenticated)/org/[organizationSlug]/teams/_components`
- [ ] `cd apps/hyperlocalise-web && vp test -- src/app/[lang]/(authenticated)/org/[organizationSlug]/teams/_components`

## Checklist

- [x] I ran relevant checks locally (`vp lint`, `vp test` on teams `_components`)

### DIFF
--- a/apps/hyperlocalise-web/src/app/[lang]/(authenticated)/org/[organizationSlug]/teams/_components/add-team-member-dialog.messages.ts
+++ b/apps/hyperlocalise-web/src/app/[lang]/(authenticated)/org/[organizationSlug]/teams/_components/add-team-member-dialog.messages.ts
@@ -1,0 +1,52 @@
+"use client";
+
+import { defineMessages } from "react-intl";
+
+export const addTeamMemberDialogMessages = defineMessages({
+  title: {
+    defaultMessage: "Add team member",
+    id: "Rnm9kBTKL/",
+    description: "Title of the add team member dialog",
+  },
+  description: {
+    defaultMessage:
+      "Assign an existing workspace member to this team. People must already belong to the workspace before they can join a team.",
+    id: "RWjNPOBQ+S",
+    description: "Description of the add team member dialog",
+  },
+  everyoneAlreadyOnTeam: {
+    defaultMessage: "Everyone in this workspace is already on the team.",
+    id: "PXpU01XAii",
+    description: "Empty state when no assignable members remain",
+  },
+  memberLabel: {
+    defaultMessage: "Member",
+    id: "3tBZU7JUcu",
+    description: "Label for the member select in the add team member dialog",
+  },
+  selectMemberPlaceholder: {
+    defaultMessage: "Select a member",
+    id: "oeVZ6ETPjK",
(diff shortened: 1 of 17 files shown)


cungminh2710:

---

### Issue 1 of 3
apps/hyperlocalise-web/src/app/[lang]/(authenticated)/org/[organizationSlug]/teams/_components/team-form.ts:1-7
**Utility module now silently depends on a client-only module**

`team-form.ts` imports `teamFormMessages` from `./team-form.messages`, which has `"use client"`. Without adding `"use client"` to `team-form.ts` itself, the file continues to look like a server-safe shared utility, but any Server Component or server-side code that transitively imports it will fail at build time with the `defineMessages() from the server` error described in `AGENTS.md`.

Currently no Server Component imports `team-form.ts` directly, so the build passes — but the silent boundary violation makes this a hidden landmine. Add `"use client"` as the first line of `team-form.ts` to make the constraint explicit.

### Issue 2 of 3
apps/hyperlocalise-web/src/app/[lang]/(authenticated)/org/[organizationSlug]/teams/_components/teams-settings-view-model.ts:1-11
**Utility module now silently depends on a client-only module**

`teams-settings-view-model.ts` now imports `teamsSettingsViewModelMessages` from `./teams-settings-view-model.messages`, which carries `"use client"`. The file itself has no directive, so it still appears server-safe — but importing it from a Server Component (or any non-client module higher in the tree) will cause the same `defineMessages() from the server` build failure documented in `AGENTS.md`. `getTeamRoleLabel` / `getTeamRoleDescription` are called from view components, but they're also re-exported utility functions that future callers might reach from a server context. Add `"use client"` as the first line to make the client-only constraint visible.

### Issue 3 of 3
apps/hyperlocalise-web/src/app/[lang]/(authenticated)/org/[organizationSlug]/teams/_components/team-form.ts:20-29
**Duplicated `resolveMessage` helper**

An identical `resolveMessage` helper (same signature, same `typeof descriptor.defaultMessage === "string"` fallback) is defined privately in both `team-form.ts` and `teams-settings-view-model.ts`. If the fallback logic ever needs to change — e.g., because `defaultMessage` can now be an object in a future `react-intl` version — it must be updated in two places. Consider extracting this into a shared utility (e.g., `src/lib/i18n/resolve-message.ts`) and importing it in both files.

PR #1493, opened 2026-07-18
fix(sandbox): install Chromium system deps for screenshot capture

cungminh2710:
## What changed

Sandbox Playwright Chromium failed to launch with a missing `libnspr4.so`. This change installs Chromium OS dependencies during sandbox bootstrap (`playwright install-deps`, with an apt fallback), retries deps during `captureScreenshot` via `--with-deps`, and maps missing shared libraries to `browser_system_deps_unavailable`. Also updates local `e2e:install` and the default Storybook screenshot viewport to 1440×1000.

## How to test

- [x] `vp test src/lib/agent-runtime/tools/workspace/capture-screenshot.test.ts src/lib/vercel-sandbox-config.test.ts src/lib/translation/sandbox-translation.test.ts src/workflows/repository-agent.test.ts` in `apps/hyperlocalise-web`
- [x] `vp check --fix` in `apps/hyperlocalise-web`
- [ ] Create a new sandbox and retry a Storybook `captureScreenshot` (e.g. App/Project/Files/Page — RepositoryFiles) and confirm the PNG artifact is stored

## Checklist

- [x] I ran relevant checks locally (`vp test` for the affected suites, `vp check --fix`)

### DIFF
--- a/apps/hyperlocalise-web/AGENTS.md
+++ b/apps/hyperlocalise-web/AGENTS.md
@@ -51,7 +51,7 @@ Fixture-auth infrastructure and baseline Playwright specs live under [`src/e2e/`
 
 - **Fixture auth**: set `E2E_AUTH_MODE=fixture` and a 32+ character `E2E_AUTH_SECRET` in `.env` (see [`.env.e2e.example`](.env.e2e.example)). Programmatic login: `POST /api/e2e/auth/session` with `X-E2E-Setup-Token`. Browser helpers in [`src/e2e/fixtures/browser.ts`](src/e2e/fixtures/browser.ts) use that API route. Disabled when `NODE_ENV=production` or `VERCEL_ENV=production`.
 - **Baseline specs**: [`src/e2e/flows/*.e2e.ts`](src/e2e/flows/) use Playwright from Node. Flows include fixture auth login, dashboard overview, project creation, and onboarding workspace creation. To run them manually, start Postgres, migrate, build, and serve the app with fixture auth, then invoke Vitest against those files directly if needed.
-- **Commands**: run `vp run e2e:install` once to install Chromium, then run `vp run test:e2e` while the fixture-auth app is available. Set `E2E_BASE_URL` when the app is not running at `http://localhost:3000`.
+- **Commands**: run `vp run e2e:install` once to install Chromium and OS libraries (`playwright install --with-deps chromium`), then run `vp run test:e2e` while the fixture-auth app is available. Set `E2E_BASE_URL` when the app is not running at `http://localhost:3000`.
 - Vitest Browser Mode is not used for route-level e2e (it cannot `page.goto` external origins).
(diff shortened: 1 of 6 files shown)


cungminh2710:

---

### Issue 1 of 2
apps/hyperlocalise-web/src/lib/vercel-sandbox-config.ts:108-110
**Bootstrap failure if Chromium deps install fails**

`install_chromium_system_dependencies` is the last statement executed in the script. In `sh` (without `set -e`), the exit code of the script is the exit code of that last `if` compound command. If the condition is true (libnspr4 is missing) and `install_chromium_system_dependencies` returns non-zero — e.g. due to a network timeout downloading the Playwright npx package, or if neither `npm` nor `apt-get` is available — the `if` compound exits non-zero, the overall `sh -c` exits non-zero, and `createConfiguredVercelSandbox` throws `"sandbox tool installation failed"`, making sandbox creation impossible. Since Chromium deps are explicitly described as a best-effort step with a capture-time retry, the failure here should be non-fatal. Add `|| true` to prevent a transient installation failure from blocking all sandbox operations.

### Issue 2 of 2
apps/hyperlocalise-web/src/lib/agent-runtime/tools/workspace/capture-screenshot.ts:503-505
**Retry overwrites original failure log**

The binary-only retry at line 504 opens `/tmp/hyperlocalise-chromium-install.log` with `>` (truncate), so if both `--with-deps` and the plain `install` both fail, `cat`-ing the log on line 505 shows only the second attempt's output. The original failure reason from `--with-deps` is silently lost, which makes debugging harder. Use `>>` on the retry to preserve the first run's output.

cungminh2710:

---

### Issue 1 of 1
apps/hyperlocalise-web/src/lib/agent-runtime/tools/workspace/capture-screenshot.ts:490-491
The generated shell line `sudo PLAYWRIGHT_BROWSERS_PATH=... playwright install-deps` does not set the env var for the sudo'd process. POSIX shell only treats `VAR=val cmd` as an env-prefix assignment when `VAR=val` is the first word of a simple command — prepending `sudo` makes `sudo` the command and `PLAYWRIGHT_BROWSERS_PATH=...` its first argument, so sudo tries to execute a binary named `PLAYWRIGHT_BROWSERS_PATH=/tmp/…` and immediately exits non-zero. `DEPS_RC` stays 1 and the privileged path is silently skipped every time. In environments where the sandbox user is not already root, `install-deps` will also fail on the no-sudo fallback, producing a `browser_system_deps_unavailable` error. The fix is `sudo env VAR=val cmd`.

```suggestion
    "  if command -v sudo >/dev/null 2>&1; then",
    `    sudo env PLAYWRIGHT_BROWSERS_PATH=${browsersPath} ${playwrightBin} install-deps chromium >/tmp/hyperlocalise-chromium-deps.log 2>&1 && DEPS_RC=0`,
```

PR #1512, opened 2026-07-20
feat(issues): add route-aware issue detail drawer

nguyenkhavi:
## Summary
- Add a URL-driven issue detail drawer on the project Issue Sheet and org Issues pages so users can inspect and edit an issue without leaving the list
- Expose `GET /issue-sheet/:issueId`, extend members API with local `userId` for assignee picking, and sync open/close via `issue` / `issueProject` search params (browser back closes the drawer; refresh restores it)
- Support editable title, description, status, type, priority, and assignee, plus Open in CAT, mobile full-screen layout, and focus return to the selected row

## Test plan
- [ ] On project Issue Sheet, click a row → drawer opens and URL includes `?issue=...`
- [ ] Browser back closes the drawer without leaving the sheet; list filters/scroll remain
- [ ] Refresh with `?issue=...` restores the drawer with loaded data
- [ ] Edit title, description, status, type, priority, assignee → changes persist after refresh
- [ ] Open in CAT navigates to the linked segment when source path + locale exist
- [ ] On org Issues, row click opens the same drawer with `issue` + `issueProject` in the URL
- [ ] Mobile viewport uses full-screen detail; keyboard focus enters drawer and returns to the row on close
- [ ] Inline table selects/inputs still edit without opening the drawer

### DIFF
--- a/apps/hyperlocalise-web/.gitignore
+++ b/apps/hyperlocalise-web/.gitignore
@@ -44,3 +44,4 @@ next-env.d.ts
 
 *storybook.log
 storybook-static
+/.swc
--- a/apps/hyperlocalise-web/src/api/routes/member/member.route.ts
+++ b/apps/hyperlocalise-web/src/api/routes/member/member.route.ts
@@ -190,6 +190,7 @@ async function inviteOrganizationMember(input: {
       isNewUser: false,
       member: toMemberSummary(
         {
+          userId: existingMembership.localUserId,
           workosUserId: existingMembership.workosUserId,
           email: existingMembership.email,
           firstName: existingMembership.firstName,
@@ -262,6 +263,7 @@ async function inviteOrganizationMember(input: {
     isNewUser,
     member: toMemberSummary(
       {
+        userId: user.id,
         workosUserId: user.workosUserId,
         email: user.email,
         firstName: user.firstName,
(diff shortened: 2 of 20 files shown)


cungminh2710:

---

### Issue 1 of 1
apps/hyperlocalise-web/src/app/[lang]/(authenticated)/org/[organizationSlug]/_components/issue-detail/issue-detail-panel.tsx:571-578
**"Saved" indicator can show while a mutation is still pending**

When both mutations have previously succeeded (`updateIssue.isSuccess = true` AND `setValue.isSuccess = true`), a new call to `updateIssue.mutate()` changes `updateIssue.isSuccess` from `true` → `false`. The effect deps `[setValue.isSuccess, updateIssue.isSuccess]` change from `[true, true]` to `[true, false]`, so the effect fires. The early-return condition `!updateIssue.isSuccess && !setValue.isSuccess` evaluates to `true && false = false`, so it is skipped, and `setShowSaved(true)` fires — even though the new mutation is still in flight. If that mutation subsequently fails, the user sees "Saved" and then an error toast side-by-side, creating a false impression that their change persisted. The cleanup also cancels the previous 2-second timeout, leaving `showSaved` stuck at `true` until the next effect run.

PR #1519, opened 2026-07-20
⚡ Bolt: optimize segment edge whitespace profile scanning

cungminh2710:
💡 What: Introduced a fast-path in `profileEdgeWhitespace` to check if the first and last bytes of a string are ASCII non-whitespace, completely bypassing rune decoding (`utf8.DecodeRuneInString` / `utf8.DecodeLastRuneInString`) and scanning. Also updated `hasProfileWhitespaceSignals` to check for non-breaking spaces using `strings.Contains` instead of a full `countNBSP` scan.

🎯 Why: Segment validation and profile verification is a hot path called for almost all translated segments. For standard segments, leading and trailing whitespace is usually absent, making the overhead of rune decoding and complete scans on these strings unnecessary and wasteful.

📊 Impact: Bypasses rune decoding for the vast majority of segments, speeding up PlainASCII edge detection to ~5.8 ns/op and 0 allocations.

🔬 Measurement: Verified with `go test -bench=BenchmarkProfileEdgeWhitespace ./internal/i18n/segmentvalidate`.

---
*PR created automatically by Jules for task [10758255464337719803](https://jules.google.com/task/10758255464337719803) started by @cungminh2710*


cungminh2710:

---

### Issue 1 of 1
internal/i18n/segmentvalidate/profile.go:83-109
The empty `if` branches with only a comment are non-idiomatic Go. Inverting the conditions makes the intent clearer and removes the need for the comment entirely.

```suggestion
	start := 0
	if !firstNonWS {
		for start < len(value) {
			r, w := utf8.DecodeRuneInString(value[start:])
			if !isProfileEdgeWhitespace(r) {
				break
			}
			start += w
		}
		leading = value[:start]
	}

	end := len(value)
	if !lastNonWS {
		for end > start {
			r, w := utf8.DecodeLastRuneInString(value[start:end])
			if !isProfileEdgeWhitespace(r) {
				break
			}
			end -= w
		}
		trailing = value[end:]
	}
```

PR #1616, opened 2026-07-29
feat(web): add per-automation Memory alongside org-wide Knowledge Memory

NguyenChHieu:
## What changed

Adds a per-automation **Memory** feature, conceptually similar to Claude's Scheduled Tasks: each workspace automation can now have its own Memory (freeform content + versioned revision history) that gets fed into the agent's prompt before it runs, alongside the existing org-wide Knowledge Memory.

- **DB**: new `workspace_automation_memories` head table + `workspace_automation_memory_revisions` archive table (migration `0069`).
- **Shared logic**: extracted the commit/revision-archive orchestration used by Knowledge Memory into `src/lib/versioned-document/`, and refactored the existing org-level Knowledge Memory service to use it (behavior-preserving — its existing 51-test regression suite still passes unchanged).
- **Service layer**: `src/lib/workspace-automation-memory/` — get/commit/list/restore, plus a plain (non-versioned) `setWorkspaceAutomationMemoryIncludeOrgKnowledge` toggle.
- **Prompt composition**: `resolveWorkspaceAutomationMemoryContext` is wired into `run-workspace-orchestrator.ts` and `compose-workspace-instructions.ts`. If an automation has its own Memory content, its `includeOrgKnowledge` checkbox becomes authoritative for whether org-wide Knowledge Memory is also included (bypassing the automation's separate `toolConfig.knowledge.enabled` gate via a new `forceInclude` param). Automations with no Memory content are unaffected — existing behavior is unchanged.
- **REST routes**: GET/PUT `/:automationId/memory`, GET `/:automationId/memory/revisions`, GET `/:automationId/memory/revisions/:revisionId`, POST `/:automationId/memory/revisions/:revisionId/restore` — ETag/If-Match optimistic concurrency, same pattern as Knowledge Memory.
- **UI**: reuses the existing "Manage" button under the automation Settings → Tools → Memories row (per product direction, this replaces the old org-wide-only editor there rather than adding a separate tab) — content editor, "Also include organization-wide Memory" toggle, and version history.
- **Availability**: gated by the same org-wide Knowledge Memory feature flag as the existing org feature (not automation-specific).

## How to test

- `vp check` — format/lint/typecheck, clean.
- `vp test run` — 3154/3163 passing. The 9 failures (5 files: `workos/app-auth.test.ts`, `onboarding/context.test.ts`, `blog/blog-post.test.ts`, `_runtime/loader.test.ts`, `lib/agents/slack/bot.test.ts`) are pre-existing and unrelated — they fail identically in isolation, due to Windows path-separator assumptions in those tests, and none touch any file in this diff.
- New tests: `workspace-automation-memory.test.ts` (5 tests), `resolve-workspace-automation-memory.test.ts` (2 tests) — both pass.
- Knowledge Memory regression suite (4 files, 51 tests) — all pass, confirming the shared-module refactor is behavior-preserving.
- Manually verified in local dev: created an automation, opened Settings → Tools → Memories → Manage, edited content, toggled "Also include organization-wide Memory", saved a version, and confirmed version history renders correctly.

## Checklist

- [x] I ran relevant checks locally (`vp check`, `vp test`)


cungminh2710:
@codex review

NguyenChHieu:
@codex review

NguyenChHieu:
Thanks for the thorough review — several of these are real and I've fixed them. I want to separate the parts I could verify from the part I couldn't, because they need different responses.

## Fixed

**Design doc drift.** Correct, and this was the most useful catch. `docs/plans/2026-07-17-automation-knowledge-memories-design.md` listed per-automation memories as out of scope and was never updated. I've amended it to record what shipped, why the scope changed, and the tri-state behaviour below.

**Manage-sheet behaviour change.** Also correct — on `main` that button opens the org `KnowledgeMemoryEditor`; on this branch it opens the per-automation one. The sheet now names which memory it edits and links back to the Knowledge page, so the org-level document stays one click away.

**`versioned-document` split.** Agreed, and done — it's out as its own behaviour-preserving PR so it can land on its own merits regardless of how the rest resolves.

**Dual gate.** Real, and now documented with a tri-state table plus tests pinning all three paths. I deliberately did **not** collapse it: `includeOrgKnowledge` defaults to `true`, so making it unconditionally authoritative would start injecting org knowledge into automations that have `knowledge.enabled === false` and no Memory — a silent behaviour change for existing users. Collapsing it properly needs a nullable column and a migration, which the open question below could make unnecessary.

## Not actioned: the Upstash direction

I checked before planning around this, and I couldn't find the grounding for it:

- **No footprint in the repo** — no `@upstash/*` dependency in any `package.json` (the only hits are inside `node_modules`, from drizzle-orm's optional cache peer), no mention anywhere in `docs/`, no `.cursor/rules`, no `BUGBOT.md`.
- **No prior decision** — no design note establishes agent-owned memory or Upstash as the target architecture.

Since the review is framed as *"the memory model we want"* and *"target direction"*, I want to flag that I can't find where that was decided. If it was settled somewhere I don't have visibility into, point me at it and I'll replan against it.

For the record on one adjacent point: earlier research into an external retrieval backend concluded it doesn't address the prompt-truncation problem it was originally raised for (that fix is #1636, and it's local). That answers a *different* question than agent-owned memory, so it isn't a rebuttal here — I mention it only so the two don't get conflated.

## The actual question

Stripped of the Upstash framing, there's a legitimate product question underneath:

> Should per-automation Memory be human-curated documents, or agent-learned facts?

Worth deciding deliberately. Two things I'd note either way:

- These aren't mutually exclusive. Agent-written memory is a plausible **addition** — several orchestrator tools already write back during a run, so there's precedent — and it doesn't require deleting the human-curated layer.
- Removing the editor, schema, and routes is a large, hard-to-reverse change. I'd rather not make it on an automated review without a human call behind it.

@cungminh2710 — this one's yours. Happy to go either way, but I'd like the decision explicit before reshaping the PR.

PR #1698, opened 2026-08-05
feat(web): let automations recall and append to org Memory via tools

NguyenChHieu:
## What changed

Gives workspace automations controlled access to org-level `Memory.md` via two tool calls, replacing the old behavior of always pasting a fixed excerpt into composed instructions. `toolConfig.knowledge` gets a second field, `allowUpdates`, surfaced as two UI toggles.

- **`recall_memory`** — agent calls this with its own query when guidance seems relevant; reuses `selectKnowledgeMemoryContext`. Re-checks the workspace-knowledge flag at call time, nonfatal if disabled.
- **`save_memory`** — appends via the same compare-and-swap path a human edit uses, so every agent append is a normal, restorable revision. Forced step, but `entry: null` is a valid "nothing to remember" outcome. Concurrent-write conflicts and the size cap are returned, not thrown.
- `commitVersionedDocument`'s `updatedByUserId` widened to `string | null` (agent commits have no human actor). `planHasActionableTool` treats a memory-only plan as not runnable.

Supersedes #1616 (per-automation store); branched fresh from `main`.

## Review round 2

4 findings from an automated pass + follow-up self-review — 1 real, 3 investigated and confirmed as false positives (not dismissed on sight):

- **Fixed**: `resolveKnowledgeMemoryEnabled` reimplemented `resolveWorkspaceKnowledgeFlag`'s own org-lookup logic — now resolves `organizationId` and delegates to the shared resolver.
- **Not a bug**: `toolConfig` throw vs. flag nonfatal looked inconsistent — traced through `plan.ts`, the `toolConfig` check is unreachable defensive code (planning already gates on it), correctly throws per this repo's own invariant-vs-expected-failure convention.
- **Not a bug**: `stepLimit` looked unbounded relative to the fixed timeout — `plannedToolCount` is capped at 11 possible tools, well within the timeout's real budget.
- **Not a bug, already covered**: "no test that recall finds guidance" — that coverage already exists in `memory-tools-integration.test.ts`.
- **P1** (found after the fix above was pushed): the dedup fix's test mocked `workspace-flags` via `importOriginal` + spread, which re-exports the *real* `resolveWorkspaceKnowledgeFlag` — its body still resolves the flag from its own module closure, so the mock override never took effect and the test silently exercised the real WorkOS adapter. Fixed by mocking the resolver function directly (same pattern already used in `memory-tools-integration.test.ts`).

## Verification

- Stash discipline on every fix; `vp check --fix` clean.
- No local Postgres in this sandbox, so ran the DB-backed suites against a standalone `@electric-sql/pglite` instance for real execution signal: `api-translation-job-queue.test.ts` 5/5, `memory-tools-integration.test.ts` 2/2 (including the real recall-finds-what-was-saved case). `workspace-automation-dispatcher.test.ts` (pre-existing, unrelated) 7/11 — the 4 failures are the same known PGlite concurrency limitation, not an app bug.
- Not yet verified against real CI — that's the one gap PGlite can't close.

## Checklist

- [x] Ran `vp check` / `vp test` locally, plus PGlite-backed execution for DB-dependent suites
- [x] Scoped and ready for review


cungminh2710:
@codex review

NguyenChHieu:
@codex review

NguyenChHieu:
@codex review

NguyenChHieu:
@codex review

NguyenChHieu:
@codex review

NguyenChHieu:
@codex review

NguyenChHieu:
@codex review

NguyenChHieu:
@codex review

NguyenChHieu:
@codex review

NguyenChHieu:
@codex review

NguyenChHieu:
@codex review

NguyenChHieu:
@codex review

NguyenChHieu:
@codex review

NguyenChHieu:
@codex review

PR #1704, opened 2026-08-05
⚡ Bolt: optimize XLIFF marshalling performance via streaming and zero-allocation tag scanning

cungminh2710:
Optimizes XLIFF translation file marshalling and rewriting performance by avoiding intermediate `xml.Token` buffering and cloning. Instead, translation units without replacements are streamed directly to the encoder. For units with replacements, boundaries are identified from the raw template, and a zero-allocation, prefix-matching-safe raw byte-level scanner identifies `<target>` tag presence while handling comments and CDATA. The replacement is then streamed on-the-fly. This results in a ~36.9% reduction in allocations and a ~32.8% reduction in memory bytes allocated, yielding a ~8.7% overall speedup.

---
*PR created automatically by Jules for task [9572266186457638366](https://jules.google.com/task/9572266186457638366) started by @cungminh2710*

---

> [!NOTE]
> **Medium Risk**
> Marshalling behavior for edge-case XML (nested units, comments, CDATA, template/decoder alignment) depends on the new streaming and raw-slice logic rather than the previous buffered token replay.
> 
> **Overview**
> **`MarshalXLIFF`** no longer buffers and clones every token inside a translation unit. Units with **no** entry in the values map are copied straight from the decoder to the encoder with a simple depth walk.
> 
> When a unit **does** have a replacement, the change slices that unit’s span from the raw template (using decoder offset and closing-tag search), then uses new **`hasXLIFFTargetElement`**—a zero-allocation byte scan that skips comments/CDATA and matches a real `<target` tag—to decide whether to rewrite **`<target>`** or **`<source>`**. Replacement text is injected by streaming tokens while skipping nested content under the chosen element, instead of replaying a cloned token buffer.
> 
> Bolt journal entry documents the streaming / raw-scan pattern for future XLIFF work.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3a1137a84e737ec0953c4c5ac517b47d62d36742. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cungminh2710:
@codex review

PR #1825, opened 2026-08-15
feat(issue-sheet): add localization issue templates

NguyenChHieu:
## What changed

Adds 5 localization issue templates (translation mistake, source mistake, context request, glossary violation, QA failure) to issue creation, replacing the blank-form / CAT-hardcoded-`context_request` behavior with structured, evidence-prompting templates.

- **Schema**: `issueSheetIssues.templateKey` (nullable, provenance-only, no FK) and `projects.issueTemplateConfig` (jsonb: default template + per-template assignee bindings), defaulting to `{}` so existing projects are unaffected until an admin opts in.
- **Templates**: 5 static definitions (key, issue type, default priority, localized skeleton headings) in `issue-sheet-templates.ts`. Each skeleton is headings-only, no prose hints, so unfilled sections get stripped structurally on submit.
- **Create dialog**: template picker with precedence (in-session pick > `initialTemplateKey` prop > project default > none), a `descriptionDirty` flag (not string diffing, since TipTap normalizes markdown) to decide when auto-applying a template is safe, and atomic apply semantics — explicit picks always overwrite, automatic (project-default) apply is all-or-nothing including the assignee binding.
- **CAT**: still forces `context_request` via `initialTemplateKey`; the create dialog now composes the description (skeleton + blockquoted source) instead of CAT dropping raw source text in directly.
- **Project settings**: new panel to set the project's default template and an optional assignee per template, with stale/rejected bindings shown and dropped rather than silently re-sent.
- **Issue detail panel**: read-only template display, hidden when no template was used.
- **API**: `GET/PUT /projects/:id/template-config` (GET open to any project member, PUT gated on `projects:write`); `POST /issues` accepts an optional `templateKey`. Update route deliberately does **not** accept `templateKey` — it's read-only provenance, non-retroactive.

This is prep for HL-503 (routing rules), which is why `templateKey` is a real column rather than inferred from `issueType`.

### A few screen shots

Screenshot 2026-08-15 101710

Screenshot 2026-08-15 101933

## How to test

```bash
cd apps/hyperlocalise-web
./node_modules/.bin/vp run dev
```

Manual walkthrough:
1. Open a project's issue sheet → **New issue** → pick a template → confirm the skeleton renders as headings and priority/type prefill.
2. Type into the description, swap templates → confirm the new skeleton replaces the old one.
3. Type into the description, then switch project in the dialog → confirm the project default does **not** clobber your text (dirty-state protection).
4. Leave some skeleton sections empty, submit → confirm empty sections are stripped from the saved description.
5. Open the create-issue dialog from CAT → confirm `context_request` is preselected and the source text appears as a blockquote below the skeleton.
6. As a project admin, go to **Project settings → Issue templates** → set a default template and an assignee for one template → save → reload → confirm it persisted.
7. Open an existing issue created from a template → confirm the read-only **Template** row shows in the issue detail panel.

Automated:
```bash
./node_modules/.bin/vp test src/lib/projects/issue-sheet/issue-sheet-templates.test.ts "src/app/[lang]/(authenticated)/org/[organizationSlug]/_components/issue-detail/issue-sheet-column-utils.test.ts" "src/app/[lang]/(authenticated)/org/[organizationSlug]/projects/[projectId]/issue-sheet/_components/issue-sheet-create-issue-dialog.test.tsx" "src/app/[lang]/(authenticated)/org/[organizationSlug]/projects/[projectId]/settings/_components/project-issue-templates-panel.test.tsx" "src/app/[lang]/(authenticated)/org/[organizationSlug]/_components/issue-detail/use-issue-detail-mutations.test.tsx" "src/app/[lang]/(authenticated)/org/[organizationSlug]/projects/[projectId]/issue-sheet/_components/issue-sheet-template-description.test.ts" "src/components/cat/issues/cat-editor-issues-section.test.tsx"
```
58/58 passing locally, plus a full manual end-to-end pass (real WorkOS sign-in, real Postgres, template picker, project settings, issue detail panel) all confirmed working.

## Checklist

- [x] I ran relevant checks locally (`vp check --fix`, `vp test` on all touched files)


cungminh2710:
@codex review

Issue #1831, opened 2026-08-15
Issue feed can misorder issue_created before a later comment under DB load


NguyenChHieu:
**What's wrong**
The Issue Sheet activity feed sorts by `created_at`, but two different clocks write that column. `issue_created` gets a JS-computed timestamp (`new Date()` in `IssueSheetService.createIssue`), while comments get the database's own `now()`. Under load, the two can disagree enough that a comment sorts before the issue-creation event that logically preceded it.

**Root cause**
`insertIssueCreatedActivity` and `insertAssigneeChangedActivity` in [issue-sheet-service.ts](apps/hyperlocalise-web/src/lib/projects/issue-sheet/issue-sheet-service.ts) take an explicit `createdAt` computed in Node, with later same-transaction activities stamped `+1ms` later to force correct relative ordering. That workaround exists because Postgres `now()` is transaction-scoped (documented, intentional Postgres behavior, verified empirically: two `now()` calls one second apart inside the same transaction returned the identical value), so two inserts in the same transaction would otherwise tie. But computing the timestamp in Node instead of the database is what introduces the clock mismatch against comments, which is the actual bug.

**Reproduces on `main`, unrelated to any in-flight feature.** Confirmed deterministic under concurrent test load, not flaky:
- `issue-sheet-feed-service.test.ts`, "interleaves activities and root comments and nests replies"
- `issue-sheet.route.test.ts`, "returns a unified feed of activities and comment threads ordered in SQL"

**Severity**
Cosmetic display-order bug only. No data loss, corruption, or business-logic impact, nothing downstream reads or depends on this ordering. Needs concurrent DB load plus a comment landing within a tight window after issue creation to manifest, so it is latent under typical usage and more likely under bulk automation or traffic spikes. Not urgent, not a release blocker.

**Fix direction**
Use `clock_timestamp()` (advances per statement, even within a transaction) for all activity inserts instead of a JS-computed timestamp, matching how comments already get their timestamp from the DB. Removes the need for the manual `+1ms` stamp entirely. Touches HL-491's original activity-insert helpers plus every feature built on top of them (HL-493, HL-501).

PR #1833, opened 2026-08-15
[HL-501] Add resolution reasons and verification to Issue Sheet

NguyenChHieu:
## What changed

Implements HL-501: resolving an issue now requires a reason (fixed / source updated / duplicate / not reproducible / won't fix), and an optional verification step gated by a designated verifier.

- **State machine** (`issue-status-transitions.ts`): `open ⇄ in_progress → {resolved | awaiting_verification | wont_fix} → verified`, reopen from any closed state back to `open`. Won't-fix is terminal and never verified. Verification is derived from whether a verifier is designated, not a separate client-authored status.
- **Backend**: schema (`resolutionReason`, `resolvedByUserId`, `verifierUserId`, `verifiedAt`, `verifiedByUserId`), migration with a backfill for legacy resolved rows, transition/permission validation reused across create and update, new activity types (resolved/verified/reopened/verifier_changed), a `verification_requested` notification, a `my_verification` list view, and summary counts split across `awaiting_verification`/`resolved`/`verified`.
- **Frontend**: resolve dialog collecting a required reason and optional verifier, a verifier field on the issue detail panel, a reopen control with an optional comment, status-dependent dropdown filtering, and feed rendering for every new activity type.
- **Review-driven hardening** (commits 3-4): fixed two P1s found in review - the resolve dialog silently dropping an already-designated verifier, and no way to un-stick an `awaiting_verification` issue once its verifier is cleared (now demotes symmetrically to `resolved`). Also: `awaiting_verification` now counts as "open work" in the general list views, a resolution reason sent without an actual close transition is rejected instead of silently ignored, `in_progress -> open` no longer misrecords as a "reopened" event, the won't-fix path no longer exposes a verifier picker, and reopening a `verified` issue is now permission-gated the same as verifying it.

Out of scope, filed separately: #1831, a pre-existing feed-ordering race unrelated to this change, found and confirmed while testing.

### Some screenshots

Screenshot 2026-08-15 101933

Screenshot 2026-08-15 224433

## How to test

```bash
vp test src/lib/projects/issue-sheet src/api/routes/project/issue-sheet.route.test.ts
vp check
```

Manually verified end-to-end against a live dev server + Postgres: reason-required validation, resolve-with-verifier -> `awaiting_verification` -> verify -> `verified`, verifier retained across a separate resolve step, verifier-clear demotion, won't-fix hiding the verifier picker, reopen with comment, `in_progress -> open` not misrecorded as reopened, summary grouping, and the "My verification" view.

## Checklist

- [x] I ran relevant checks locally (`vp check`, `vp test`)


cungminh2710:
I don't think we need a verifier, we record who resolved the issue in the activity log

cungminh2710:
@codex review

PR #1865, opened 2026-08-16
feat(issue-sheet): add issue relationships and duplicate handling

NguyenChHieu:
## What changed

Adds relationship tracking between issues in the Issue Sheet: `related`, `blocks`, `blocked_by`, and `duplicate_of`, visible in the issue detail panel, with a searchable cross-project target picker, circular-chain rejection, and non-destructive removal. Relationship changes show up in the issue's activity feed.

Built across: schema + circular-relationship guard, access-scoped service layer with cross-project support, API routes + org-wide search for the target picker, detail-panel UI, and activity feed entries.

Two things caught and fixed during self-review/testing, not part of the original design:
- **Security**: a route checked project access but never verified the `issueId` in the URL actually belonged to that `projectId` — cross-team IDOR. Fixed with a fail-before/pass-after regression test.
- **UI**: the relationship-kind picker rendered the raw enum (`related`, `blocked_by`) instead of a translated label — a missing `label`/`items` prop on the shared `Select` component, the only place in the codebase missing that convention. Fixed, consolidated three divergent label sets into one shared module, and added per-kind icons to group headings and the activity feed (previously avatar-only). Also fixed the relationships section asserting "No relationships yet" while still loading, and made the feed's removal copy kind-specific instead of one generic sentence for all four kinds.

## How to test

```bash
vp check --fix
vp test
```

Manually verified end-to-end in a real browser: created issues, linked all four kinds both directions, verified cross-relationship inverse labels, verified 3-hop cycle rejection (a→b→d→a correctly rejected), verified duplicate banner, verified non-destructive removal with live feed copy for two of the five removal-kind variants (the other three are TypeScript-exhaustiveness-checked against a locked-down write path, not individually eyeballed).

## Checklist

- [x] I ran relevant checks locally (`vp check --fix`, `vp test`)


cungminh2710:
@codex review

PR #1869, opened 2026-08-16
⚡ Bolt: optimize extra placeholder extraction in segment validation

cungminh2710:
💡 What: Refactored `extractExtraPlaceholders` in `internal/i18n/segmentvalidate/extra_placeholders.go` to use an iterative `strings.IndexAny` + `FindStringIndex` scanning loop instead of `FindAllStringIndex`.
🎯 Why: `FindAllStringIndex` allocates a 2D slice (`[][]int`) for all matches across the string up front and scans every character using the regex engine.
📊 Impact:
- `BenchmarkValidateSegment/PrintfPlaceholders`: speedup of ~53.3% (from 7,261 ns/op down to 3,391 ns/op).
- Memory allocations: reduced by ~54.4% (from 882 B/op down to 402 B/op) and allocations reduced from 11 allocs/op to 9 allocs/op.
🔬 Measurement: Verified with `go test -bench=BenchmarkValidateSegment -benchmem ./internal/i18n/segmentvalidate`.

---
*PR created automatically by Jules for task [3800220987772680540](https://jules.google.com/task/3800220987772680540) started by @cungminh2710*


cungminh2710:
@codex review

PR #1939, opened 2026-08-20
feat(glossary): add terms during concept creation

nguyenkhavi:
## Summary
- allow adding multiple glossary terms while creating a concept
- persist the concept and additional terms atomically
- reject duplicate locale/text terms
- add route and Storybook coverage

## Validation
- `vp check --fix` passes with four pre-existing unrelated warnings
- `vp test` could not complete because PostgreSQL was unavailable at `localhost:5432`


cungminh2710:
@codex review

PR #2006, opened 2026-08-24
fix(agent-runtime): block symlink roots in grep and glob

cungminh2710:
## What changed

- Add a shared guarded search executor that rejects symbolic-link search roots and parents before running `rg`, `grep`, or `find`.
- Reject workspace-escaping paths instead of silently searching the repository root.
- Add `--no-follow` to glob’s ripgrep invocation.
- Treat flag-like grep patterns as search text so values such as `--follow` cannot override safety flags.
- Add regression coverage for symlinks targeting locations both inside and outside the workspace.

## How to test

- `vp test src/lib/agent-runtime/tools/repo-read-tools.test.ts src/lib/agent-runtime/tools/workspace/glob.test.ts` — 93 tests passed.
- `vp check --fix` — passed with nine unrelated existing warnings.
- `vp test` — attempted, but database-backed coverage could not complete because the local Drizzle migration exited with code 1.

## Checklist


cungminh2710:
@codex review

PR #2023, opened 2026-08-25
fix(web): guard marketing SSR and add SEO metadata

cungminh2710:
## Summary
- render the browser-dependent CAT demo client-side to prevent localStorage access during SSR
- add Organization JSON-LD across public marketing pages
- emit index/follow robots metadata for marketing routes and noindex/nofollow elsewhere by default

## Verification
- vp test src/components/marketing/hero-frame-mesh-stage.test.tsx src/components/seo/organization-json-ld.test.ts src/lib/seo/robots-metadata.test.ts
- vp check --fix (0 errors; 9 pre-existing unrelated warnings)

## Environment note
- the full vp test suite was attempted, but this checkout has no .env or DATABASE_URL; database-backed tests could not initialize and localhost connections are sandbox-restricted

### DIFF
--- a/apps/hyperlocalise-web/src/app/[lang]/(marketing)/layout.tsx
+++ b/apps/hyperlocalise-web/src/app/[lang]/(marketing)/layout.tsx
@@ -10,11 +10,22 @@
  * of this software will be governed by the GNU General Public License
  * Version 2.0 or later.
  */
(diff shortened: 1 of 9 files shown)


cungminh2710:
@codex review

PR #2028, opened 2026-08-25
feat(cat): add glossary guidance panel

nguyenkhavi:
## Summary
- add concordance-backed glossary concept cards with term statuses, metadata, copy actions, and glossary links
- add a floating Glossary guidance panel opened from the CAT app-shell footer
- add preferred/not-recommended match signals and Storybook coverage for results, loading, and empty states

## Validation
- `vp check --fix` passes with 0 errors and 9 pre-existing warnings
- focused tests pass: 2 files, 15 tests
- `vp test` completed with 4,602 passing tests, but 415 failures caused by the local database schema being behind the current code (`projects.identifier` is missing), plus local `localhost:3000` connection refusals

### DIFF
--- a/apps/hyperlocalise-web/src/components/app-shell/app-shell-client.tsx
+++ b/apps/hyperlocalise-web/src/components/app-shell/app-shell-client.tsx
@@ -14,6 +14,7 @@
  */
 import type { CSSProperties, ReactNode } from "react";
 import Image from "next/image";
+import { usePathname } from "next/navigation";
 import { FormattedMessage, useIntl } from "react-intl";
 
 import {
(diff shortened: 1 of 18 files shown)


cungminh2710:
@nguyenkhavi can you give me a screenshot? I thought the existing editor is showing glossary already?

nguyenkhavi:
> @nguyenkhavi can you give me a screenshot? I thought the existing editor is showing glossary already?

The existing editor already shows glossary matches, but only in a `term-based` format—for example, `“Reseller” → “Đại lý.”`

I’ve updated it to use a `concept-based` format, following Crowdin’s UI, so translators and interpreters can see more useful context, such as definitions, preferred and discouraged terms, part of speech, and grammatical information.

The main goal is to normalize our schema against Crowdin’s glossary model first. After that, I can build an adapter for native glossaries.

As part of the UI/UX update, Glossary Guidance has also been moved into a dedicated panel, which can be opened from the Glossary Guidance indicator in the app shell footer.

Screenshot @cungminh2710 
image

cungminh2710 (approved):

PR #2097, opened 2026-08-28
fix(deploy): call go-svc over the internal service binding

cungminh2710:
## Summary
- Stop exposing go-svc through a public Vercel rewrite. Next.js now proxies `/api/go-svc` to the injected `GO_SVC_URL` binding and forwards the WorkOS session cookie.
- go-svc only serves native `/health` and `/v1/...` paths; the leftover `/api/go-svc` prefix shim is gone.
- CAT still posts to the same-origin `/api/go-svc/v1/validate/segment` URL.

## Test plan
- [ ] Confirm a CAT segment validation request hits Next.js, then go-svc over `GO_SVC_URL`, and returns checks while signed in
- [ ] Confirm an unauthenticated `/api/go-svc/v1/validate/segment` call still returns 401 from go-svc
- [ ] Confirm `/api/go-svc/health` succeeds locally when go-svc is on `:8080`
- [ ] Confirm public `/api/go-svc` no longer routes around Next.js (no direct container rewrite)
- [ ] Smoke a normal page load to confirm the web catch-all rewrite still serves Next.js

### DIFF
--- a/apps/go-svc/handler.go
+++ b/apps/go-svc/handler.go
@@ -40,27 +40,10 @@ func newHandler() *handler {
 	}
 }
 
-const publicPathPrefix = "/api/go-svc"
-
 func registerRoutes(mux *http.ServeMux, h *handler, verifier SessionVerifier) {
 	validate := authMiddleware(verifier)(http.HandlerFunc(h.validateSegment))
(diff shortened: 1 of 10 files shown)


cungminh2710:
@codex review

PR #2098, opened 2026-08-28
feat(sandbox-image): add hunspell and pinned dictionaries; ADR for CLI spell check

cungminh2710:
## What changed

Three things, in increasing order of scope.

**1. Sandbox image ships Hunspell and the 20 pinned dictionaries.** Agents running in the sandbox had no spell checker. The image now installs the `hunspell` binary and stages the locales from [`internal/i18n/spellcheck/DICTIONARIES.md`](../blob/main/internal/i18n/spellcheck/DICTIONARIES.md) using the same [`apps/go-svc/build/fetch-dictionaries.sh`](../blob/main/apps/go-svc/build/fetch-dictionaries.sh) script that `Dockerfile.vercel` runs for `go-svc`, so both images carry byte-identical dictionaries verified against pinned upstream commits and SHA-256 checksums.

The build context moves from `apps/sandbox-image` to the repository root so the new first stage can reach the manifest and the script. The existing root `.dockerignore` keeps the transferred context at **16.8 kB** in CI. Licence evidence ships to `/usr/share/doc/hunspell-dictionaries/` because the dictionaries are redistributed under GPL/LGPL/MPL terms. `DICPATH` is derived from `HUNSPELL_DICT_DIR` so `hunspell -d en_US` still resolves when the image is built with a non-default dictionary path.

Local push commands in the README now run from the repository root and pass `--file apps/sandbox-image/Dockerfile` (Codex review).

**2. Fixed a real spell-check bug this uncovered (affects `go-svc` today).** Hunspell reads the first line of a `.dic` as the word count and loads an **empty** word list when it cannot parse one — silently reporting every word in that locale as misspelled. `ms_MY` ships `#30975`.

The follow-up is now defense in depth, not only an image rewrite:

- `fetch-dictionaries.sh` strips a leading `#` and surrounding whitespace, rewrites the line to a bare integer (also handles `# 30975` and a trailing comment), and fails the build when no digits remain. A `--normalize-header` helper is covered by `fetch-dictionaries_test.sh`.
- `hunspell.New` applies the same rules at load time. A `#`-prefixed header now loads; a non-numeric header fails the load so the locale is **skipped** instead of flagging every word. That takes effect on the next `go-svc` deploy even if a raw upstream file is still on disk.

**3. ADR recommending how the CLI should spell check.** [`docs/adr/2026-08-28-cli-spellcheck-design.md`](../blob/cursor/sandbox-image-hunspell-c3a7/docs/adr/2026-08-28-cli-spellcheck-design.md) recommends shelling out to `hunspell -a` (pipe mode) behind the same provider seam `go-svc` uses, skipping the check when the binary or dictionary is missing. No implementation in this PR — the ADR settles the approach.

## How to test

```bash
# Build from the repository root (the build context is now the root).
docker build --platform linux/amd64 \
  -f apps/sandbox-image/Dockerfile -t hyperlocalise-sandbox:local .

docker run --rm hyperlocalise-sandbox:local printenv DICPATH
# /usr/share/hunspell

docker run --rm -v "$PWD/apps/sandbox-image/verify-hunspell.sh:/verify.sh:ro" \
  hyperlocalise-sandbox:local bash /verify.sh

bash apps/go-svc/build/fetch-dictionaries_test.sh
go test ./apps/go-svc/internal/hunspell/
CGO_ENABLED=1 go test -tags cgo_hunspell ./apps/go-svc/...
```

`verify-hunspell.sh` checks every locale: readable `.aff`/`.dic`, Hunspell reporting it loaded, a non-empty word list, and SHA-256 match against the staged set. Result: **20 verified, 0 failed**.

Other checks: `make fmt` (clean), `make lint` (0 issues), `go test ./...` (pass), `shellcheck` on both shell scripts (clean). `make test` fails only on the pre-existing `go: no such tool "covdata"` issue documented in `AGENTS.md`.

## Notes for review

- **apt `hunspell` depends on `hunspell-en-us`**, which ships its own `en_US` into `/usr/share/hunspell`. The pinned set wins because `COPY` runs after apt; `SHA256SUMS` makes a reorder fail the build.
- Four dictionaries declare legacy 8-bit encodings. The `hunspell` binary reads `SET` and transcodes internally (`-i UTF-8`). The CGO wrapper still transcodes itself.
- Dictionary basenames do not always follow the BCP 47 tag. Resolve through `spellcheck.LoadRegistry()`.

## Checklist

<div><a href="https://cursor.com/agents/bc-01a04a9f-453e-7d36-b345-6f06d956c3a7?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-01a04a9f-453e-7d36-b345-6f06d956c3a7&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.github/workflows/sandbox-image.yml
+++ b/.github/workflows/sandbox-image.yml
@@ -5,12 +5,16 @@ on:
     types: [opened, synchronize, reopened, ready_for_review]
     paths:
       - apps/sandbox-image/**
+      - apps/go-svc/build/fetch-dictionaries.sh
+      - internal/i18n/spellcheck/DICTIONARIES.md
       - .github/workflows/sandbox-image.yml
   push:
     branches:
       - main
     paths:
       - apps/sandbox-image/**
+      - apps/go-svc/build/fetch-dictionaries.sh
+      - internal/i18n/spellcheck/DICTIONARIES.md
       - .github/workflows/sandbox-image.yml
   workflow_dispatch:
 
@@ -100,7 +104,9 @@ jobs:
       - name: Build and push
         uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6
         with:
-          context: apps/sandbox-image
+          # Repository root: the dictionary stage stages files from
+          # internal/i18n/spellcheck and apps/go-svc/build.
+          context: .
           file: apps/sandbox-image/Dockerfile
           platforms: linux/amd64
           # PRs: build-only with a local tag (no registry auth).
--- a/apps/go-svc/build/fetch-dictionaries.sh
+++ b/apps/go-svc/build/fetch-dictionaries.sh
@@ -113,6 +113,23 @@ while IFS=$'\t' read -r bcp47 aff dic evidence_csv repo; do
     cp "$aff_src" "$STAGE_DICT/$aff"
     cp "$dic_src" "$STAGE_DICT/$dic"
 
+    # Hunspell reads the first .dic line as the word count and loads an empty
+    # word list when it cannot parse one, which silently flags every word in
+    # that locale. ms_MY ships "#30975"; drop a leading '#' when the rest is a
+    # bare integer, then require a numeric header so a future dictionary with a
+    # different malformed count fails the build instead of shipping empty.
+    header="$(head -n1 "$STAGE_DICT/$dic" | sed -e '1s/^\xEF\xBB\xBF//' | tr -d '\r')"
+    if [[ "$header" =~ ^#([0-9]+)$ ]]; then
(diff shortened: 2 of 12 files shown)


cungminh2710:
@codex review

PR #2272, opened 2026-09-07
fix(phrase): continue filtered CAT pagination scans

cungminh2710:
## What changed

Phrase CAT filtered queues could incorrectly report the end of the queue when the client-side scan budget was reached before all matching keys were examined.

This change returns a continuation cursor and `hasMore: true` so subsequent requests resume scanning from the next Phrase page. Added regression coverage for matches found beyond the scan budget and documented the behavior in an ADR.

## How to test

- `vp test src/lib/providers/adapters/phrase/phrase-provider-live-content-editor-core.test.ts` — 11 passed.
- `vp check --fix` — formatting completed; existing unrelated dependency and lint issues remain.
- Full `vp test` was attempted; integration failures were caused by sandbox `EPERM` connections to local Postgres and port 3000.

## Checklist


cungminh2710:
@codex review

PR #2313, opened 2026-09-09
feat(memory): support Crowdin CSV imports

nguyenkhavi:
## Summary
- support Crowdin translation-memory CSV exports with a locale header row
- preserve the existing generic CSV import format
- add parser and route coverage for valid and malformed Crowdin rows

## Validation
- `vp test run src/lib/memory/import-memory-entries.csv.test.ts src/api/routes/memory/memory.route.test.ts` (9 tests passed)
- `vp check --fix` passed with 9 pre-existing warnings
- full `vp test` was not completed because unrelated auth/MCP suites were failing in the local environment

### DIFF
--- a/apps/hyperlocalise-web/src/api/routes/memory/memory.route.test.ts
+++ b/apps/hyperlocalise-web/src/api/routes/memory/memory.route.test.ts
@@ -190,6 +190,59 @@ describe("memoryRoutes", () => {
     ]);
   });
 
+  it("imports Crowdin's locale-header CSV export", async () => {
+    const { identity, memory } = await fixture.createStoredMemoryFixture();
+    const headers = await fixture.authHeadersFor(identity);
+
+    const response = await client.api.orgs[":organizationSlug"]["translation-memories"][
+      ":memoryId"
+    ].entries.import.$post(
+      {
+        param: {
+          organizationSlug: identity.organization.slug ?? "missing-slug",
+          memoryId: memory.id,
+        },
+        json: {
+          format: "csv",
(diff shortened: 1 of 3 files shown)


nguyenkhavi:
@codex review

nguyenkhavi:
Addressed the Codex review feedback in commit `3a20c135`:

- Crowdin CSVs with multiple target locales now expand each target column into its own TM entry.
- Crowdin format detection now requires locale-shaped headers and matching row width, preserving generic CSV imports after malformed two-cell preambles.
- Added coverage for multi-locale exports and the generic CSV fallback.

Validation: 11 focused tests passed; `vp check --fix` passed with only the existing 9 warnings.

nguyenkhavi:
Addressed the Greptile findings in commit `00b84a19`:

- Blank target cells no longer discard other translated locales from the same Crowdin row.
- Locale-header detection now validates language names via `Intl.DisplayNames`, avoiding false positives such as `OK` in headerless generic CSVs.
- Added regression tests for partial rows and the generic `en,fr,OK,OK` case.

Validation: 13 focused tests passed; `vp check --fix` passed with only the existing 9 warnings.

nguyenkhavi:
Addressed the latest valid Greptile finding in commit `539c8380`:

- Crowdin locale headers using underscores (for example, `en_US,fr_FR`) are now normalized for validation and imported successfully.
- Added a regression test.

Validation: 14 focused tests passed; `vp check --fix` passed with only the existing 9 warnings. The indentation suggestion was not applied because it conflicts with the repository formatter and existing TypeScript style.

cungminh2710 (approved):

PR #2321, opened 2026-09-09
fix(reporting): skip capture when workspace reports is off

cungminh2710:
## What changed

`workspace-reports` already hid Reports in nav and the API, but translation jobs still ran `captureFileAnalysisStep` (per-string TM matching) on the critical path. Capture now checks the same flag: file jobs skip those steps when it is off, and analysis/completions/status/AI usage capture no-op for string jobs and CAT edits too.

Prefill (TM reuse and existing project translations) is unchanged. That is what `hl run` needs.

## How to test

- [ ] Run a file translation job on a workspace with `workspace-reports` off. The workflow should go extract → TM/prefill → `runTranslationStep` without `captureFileAnalysisStep` spans.
- [ ] Run the same job with the flag on. Analysis capture still runs per target locale.
- [ ] `vp test src/lib/reporting/capture-reporting.test.ts` from `apps/hyperlocalise-web`

## Checklist

- [x] I ran relevant checks locally (`vp check --fix`; `capture-reporting.test.ts` passed)

### DIFF
--- a/apps/hyperlocalise-web/src/lib/flags/workspace-flags.ts
+++ b/apps/hyperlocalise-web/src/lib/flags/workspace-flags.ts
@@ -180,23 +180,13 @@ export async function resolveWorkspaceVisualMockFlag(input: {
   }
 }
 
-/**
- * Resolves workspaceKnowledgeFlag from just an internal organizationId, for callers with no live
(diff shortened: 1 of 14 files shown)


cungminh2710:
@cursoragent fix ci and update the pr description to match the fixes we added

cungminh2710:
@cursoragent address comments

PR #2335, opened 2026-09-10
feat(glossary): connect HL-641 terminology management

nguyenkhavi:
## Summary

Implements the first end-to-end slice of HL-641 for native glossaries:

- Connects the glossary detail UI to the server-backed `/concepts/page` API.
- Adds server-side search across concept and term content, locale and review-status filters, archived visibility, stable sorting, totals, and cursor pagination.
- Loads the selected concept independently so the index no longer needs to fetch the full glossary collection.
- Fixes boolean query parsing so explicit `false` values remain false.
- Separates concept review status from term review status and excludes archived terms consistently.
- Prevents provider-backed glossary IDs from being written to native UUID history columns.
- Records before/after history for native concept and term create, update, and delete mutations.
- Adds a dedicated glossary history page with API-backed search and event-type filters, plus links from glossary detail and history events back to concepts.
- Adds regression coverage for query parsing and review filters.

## Validation

- `vp check --fix` passes with no lint or type errors.
- Focused glossary schema tests pass: 2/2 in the latest run; earlier focused schema and policy validation was 20/20.
- `git diff --check` passes.

The full suite currently reports 927 passing tests and 19 failing files because the local database is missing migrated tables and columns, including `review_status`; this is an environment migration issue rather than a UI type-check failure. The glossary route suite has the same local migration blocker before native concept mutations execute.

## Scope and follow-up

This PR establishes browse/search/pagination integration and the dedicated terminology history experience. The remaining HL-641 work is review queues and transition actions, comments/reasons, complete import/sync history, bulk operations, duplicate detection, merge/redirect workflows, restore/delete authorization, term-level pagination, and browser/API coverage for those flows.


nguyenkhavi:
@codex review

nguyenkhavi:
Addressed all Codex review findings in commit `118e4cbd`:

- History events are recorded before concept/term deletion so foreign-key constraints do not turn successful deletes into 500 responses.
- Existing concept pages now render terms from the independently fetched concept.
- Live Crowdin glossary lists use the supported concepts endpoint again.
- The list sort control is now labeled Last modified to match the API ordering.
- Paginated concept summaries include definitions and the UI renders them.

Validation: `vp check --fix`, focused glossary schema tests (2/2), and `git diff --check` pass.

nguyenkhavi:
Addressed the Greptile findings in commit `96614e7b`:

- Moved native concept/term history writes into the same database transactions as create, update, and delete mutations.
- Removed live-row foreign keys from immutable history concept/term IDs so hard deletion preserves audit identifiers and filters.
- Restored filtered export locale derivation from the active language filter.
- Added explicit history API error rendering instead of showing a misleading empty state.

Validation: `vp check --fix`, focused glossary schema tests (2/2), `git diff --check`, and generated migration `0118_medical_sandman.sql`.

cungminh2710:
@codex review

nguyenkhavi:
Addressed the latest Codex follow-up findings in commit `4875022f`:

- Reset concept pagination cursor when changing sort direction.
- Added offset pagination and previous/next controls to glossary history, using API pagination metadata.
- Persist authenticated creator IDs on native concept and term creation paths so `createdByUserId` filters work.

Validation: `vp check --fix`, focused glossary schema tests (2/2), and `git diff --check` pass.

nguyenkhavi:
Validated this finding against the current branch. It is fixed in commit `96614e7b`:

- Native concept and term create/update/delete methods now write their history rows through the same `db.transaction` as the glossary mutation.
- History writes use the transaction client, so a failed audit insert rolls back the content mutation.
- The history concept/term IDs are immutable audit identifiers without live-row foreign keys, preserving deletion history safely.

`vp check --fix` and `git diff --check` pass.

nguyenkhavi:
Fixed 2 remaining P2 issues @cungminh2710

nguyenkhavi:
Archived as the reference implementation for HL-641. This PR will be replaced by smaller independently deployable phases, starting with paginated native glossary management.

nguyenkhavi:
Closing this PR for reference. Replacement phased PRs will be opened from main.

PR #2394, opened 2026-09-13
fix(ingest): remove stale translation keys on source uploads

cungminh2710:
## What changed

Source uploads now reconcile the complete key set within the existing Vercel ingest workflow. Removed keys and their translations are deleted atomically, while retained keys keep their IDs and translations.

- Handle empty files, blank source values, and imports exceeding 5,000 keys.
- Guard against retries and out-of-order workflows.
- Reconcile repeated historical hashes so reverted uploads work correctly.
- Add a migration for the applied-version marker and QA reference indexes.

Existing stale keys are cleaned up on the next file upload.

## How to test

- `vp check --fix` — passed.
- `vp test --testTimeout=20000` — all 7,189 tests passed using isolated PostgreSQL and GNU coreutils.
- `vp run db:generate` and `vp run db:migrate` — migration generated and applied successfully to the test database.
- Regression coverage includes removals, retained translations, file isolation, empty files, large imports, stale workflows, retries, and transaction rollback.

## Checklist

- [x] I ran relevant checks locally (`vp check --fix`, `vp test`; no Go changes)


cungminh2710:
@codex review

PR #2396, opened 2026-09-13
feat(storage): add Go storage drivers and guideline search

cungminh2710:
## What changed

Add Go storage foundations for uploaded files, translation bundle distribution, and guideline retrieval.

- Support S3 and Cloudflare R2 through a shared driver, with multiple named locations configured using explicit environment variables. Object references retain their location so changing the write default does not redirect existing reads.
- Add streaming object operations, signed uploads/downloads, and immutable bundle publication with verified content and a manifest written last.
- Add an optional turbopuffer guideline index backed by canonical PostgreSQL revisions. Retrieval rejects stale or out-of-scope hits and preserves mandatory instructions when search is unavailable.
- Register storage and guideline routes under `/v1` in the existing go-svc handler. Reuse the existing WorkOS session and server-call authentication without introducing another service or secret.
- Document configuration, provider boundaries, and adoption steps.

This PR adds opt-in Go capabilities. Existing TypeScript Vercel storage and lexical guideline retrieval remain unchanged. Web migration, transactional indexing workers, release-channel APIs, and CDN configuration are follow-up work.

## How to test

Passed locally:

- `make fmt`
- `make lint`
- `TMPDIR=/private/tmp make test` — uses a canonical macOS temporary path to avoid an existing CLI symlink-path assertion failure.
- `go test -race ./internal/objectstore/... ./internal/distribution ./internal/guidelines/... ./apps/go-svc`
- `go test -race ./apps/go-svc` after the final route and configuration changes.

Tests cover parallel provider configuration, location routing, conditional writes, signed requests, publication retries, revision/scope filtering, mandatory guideline fallback, and shared authentication. Provider HTTP behavior uses local test servers; live cloud accounts were not exercised.

## Checklist


cungminh2710:
@codex review

PR #2434, opened 2026-09-16
fix(visual-workflows): preserve Switch and For Each handles on quick-add

NguyenChHieu:
## Summary
- Quick-add and drag-connect now keep the selected execution handle for Switch (`0`/`1`/`default`) and For Each (`each`/`done`) instead of dropping it to `null`.
- Adding from For Each **Each item** updates `bodyNodeIds`; deleting a body node prunes stale membership.
- Shared `execution-handles` list is used by compile and the editor. Invalid handles are rejected rather than remapped.
- Per-output `+` on multi-output nodes, plus a seeded Storybook play that asserts Case 2 / Each / Done edges and an empty validation issue panel.

Closes HL-745.

## Test plan
- [ ] `cd apps/hyperlocalise-web`
- [ ] `vp test src/lib/visual-workflows/validation/execution-handles.test.ts src/lib/visual-workflows/editor/graph-connection.test.ts src/lib/visual-workflows/visual-workflow.test.ts src/lib/visual-workflows/runtime/production-contracts.test.ts`
- [ ] `vp check --fix` on the touched files (do not run bare)
- [ ] Storybook: **App / Automations / Visual Workflow / Quick-add Switch and For Each** — play should pass (Case 2, Each item, Done)
- [ ] Manual: quick-add from If true/false still works; save/reload keeps the selected Switch/For Each route


cungminh2710:
@codex review

